PR #3, opened 2020-05-22
Add fmt and misc to readme

pepe:
I put it at the top, so when we direct someone to the correct Janet format, it is the first thing she sees in the README.


pepe:
I added dedent to README too. So it is complete.

bakpakin:
Thanks, looks good to me. The readme could afford to have more detail :)

PR #4, opened 2020-05-29
Remove printing of logs

pepe:
Please take this as the start of the debate, if prints should be part of this library. It is nothing critical, but they mangle test log output of one of my library :-).


bakpakin:
I don't have anything against dropping most of the printing, since most of it can be added in the `env` argument. The one thing you can't currently no is when a client closes, but I would say that is of minimal concern right now.

Issue #5, opened 2020-05-29
Test suite


pepe:
Are you considering the addition of the test suite? If so, may I try to do it?

I will use tester helper from your other libraries, and I will try to follow your test philosophy.

bakpakin:
Feel free to add tests here - There aren't many simply because I haven't gotten to it. Tests are always welcome!

pepe:
Thank you! I will try.

pepe:
So I added first two (easier ones) suites, please can you check if it LGTY?

pepe:
Closed in #6

PR #6, opened 2020-06-25
Add test suite to fmt, misc, msg and rpc.

pepe:
I have only implemented tests for fmt and misc, as low hanging fruits. Requesting a direction approval :-)

As debated in #5

### DIFF
--- a/test/helper.janet
+++ b/test/helper.janet
@@ -1,0 +1,47 @@
+# Helper code for running tests
+
+(var num-tests-passed 0)
+(var num-tests-run 0)
+(var suite-num 0)
+(var numchecks 0)
+(var start-time 0)
+
+(defn assert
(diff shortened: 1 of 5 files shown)


bakpakin:
As a start, this looks good to me. Obviously, it’s a bit less obvious how to test networking utilities but first things first.

pepe:
Exactly my way of thinking. I will proceed with it. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

pepe:
An only right way I have found is to use thread for the "server". I had to add some os/sleep to wait on server thread to start. 

How to test netrepl I do not know :-).

pepe:
Closes #5

pepe:
Thank you very much!

PR #11, opened 2021-01-10
Working versions of misc/timeit and misc/set*

MikeBeller:
This commit adds two functions to spork/misc -- timeit, a macro for timing pieces of code using os/clock, and set*, a macro for parallel set.  Both of these were discussed with @bakpakin on gitter today so I didn't create an issue before submitting the PR.  Caveat -- I'm not a macro expert so, regardless of the fact that I did provide some working tests, you should feel free to reject based on style or any concerns you have.  Just let me know what I would need to do to clean it up.

### DIFF
--- a/spork/misc.janet
+++ b/spork/misc.janet
@@ -25,3 +25,34 @@
         y))
     x))
 
+(defmacro timeit
+  ```Time the execution of `form` using `os/clock` before and after,
+  and print the result to stdout.  returns: result of executing `form`.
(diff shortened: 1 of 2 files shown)


bakpakin:
Thanks, @MikeBeller, the macros look good to me, merging. They look simple enough even without testing.

MikeBeller:
Thanks @bakpakin  -- I just realized -- should I add something to README in the "misc" section also?

PR #15, opened 2021-01-18
Add simple test helper, extracted from janet-lang

pepe:
I am using this helper, extracted from other janet-lang libraries, in some of my libs, usually with spork already there, so I think it could be a good idea to have it here.

I am also thinking about moving `timeit` fn from `misc` and extract `capture-stdin` into this module. What do you think, @MikeBeller & @bakpakin?


spork/test/init.janet
```diff
@@ -25,21 +25,25 @@
   x)
 
 (defmacro assert-error
+  "Test passes if forms error."
   [msg & forms]
   (def errsym (keyword (gensym)))
   ~(,assert (= ,errsym (try (do ,;forms) ([_] ,errsym))) ,msg))
 
 (defmacro assert-no-error
+  "Test passes if forms not error."
```
sogaiu, line 34: How about:

> Test passes if forms do not error.

?

pepe: yep

pepe:
Funny thing, I have not added tests for this module. I will do it later today.

MikeBeller:
Yes I think adding the very basic testing capability to spork is a very good idea.  And if you do that it makes sense to move timeit in there, and add capture-stdout.  I really didn't test capture-stdout well since it was a one-off inside a test suite so if you are going to add it maybe add a few tests for it too?

pepe:
Yep, I will add tests for the whole test module.

pepe:
So I added tests for `spork/test` and made all the other tests use `spork/test`, which reminded me about Uroboros.

I also removed the old `test/helper.janet` after extracting the `suppress-stdout` macro from it.

bakpakin:
Makes sense, that bit of code was being copied around a lot...

LGTM.

PR #21, opened 2021-01-20
Add error if there is anything after :root

### LINKED ISSUE #14: `fmt/` removes code for some cases
Easy repro:

```
➜ janet -l spork/fmt
Janet 1.13.2-dev-local linux/x64 - '(doc)' for help
repl:1:> (def c "(def a 1)\n(def b 2)\ndef c (+ a b))\n(print c)")
"(def a 1)\n(def b 2)\ndef c (+ a b))\n(print c)"
repl:2:> (format-print c)
(def a 1)
(def b 2)
def c (+ a b)
nil
repl:3:>
```

As could be seen, the `format-print` unexpectedly removes the last line and paren on the line before. I think error would be better here.

pepe:
It fixes #14 by throwing an error if there is code after `:root` in `:main`.


pepe:
The truth is this solution was @sogaiu's idea.

PR #22, opened 2021-01-21
Add temple

pepe:
I moved the better half of the Temple's test to examples/temple/, the basic.janet, and templates I suited to test/suite8.janet. There I made one code change.

I put the whole readme before the Test part under the Temple heading.

The `temple/temple.janet` became `spork/temple.janet`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,46 @@ arguments to `argparse`.
 
 Run `(doc argparse/argparse)` after importing for more information.
 
+# Temple
+
+HTML templates for Janet.
+
+Simplified version of Mendoza's template system that is cleaner and easier to use.
+Templates can be used recursively, and output is printed via `print`, so goes to
+`(dyn :out)`.
+
+Expands on the mendoza templates with the `{-` ... `-}` brackets, which do non-escaped
+substitution, so temple can be used for formats besides HTML.
+Also exposes the `escape` function inside templates for HTML escaping
+if you want to manually print to template output.
+
+## Example
+
+### foo.temple
+
+```
+{$ (def n 20) # Run at template compile time $}
+<html>
+  <body>
+    {{ (string/repeat "<>" n) # HTML escaped }}
+    <ul>
+      {% (each x (range n) (print "<li>" x " " (args :a) "</li>")) # No auto-print %}
+    </ul>
+    {- (string/repeat "<span>1</span>" n) # Not HTML escaped -}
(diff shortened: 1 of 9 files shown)


MikeBeller:
How do you decide what's in the "standard library" (if that's how we view spork)?   100% in favor of basic testing utilties, etc being in spork.  But HTML templates?  I'm not so sure.  There are other approaches to html rendering that are in joy and janet-html.  (Not templates exactly.)  I definitely don't feel strongly about this either way but wanted to at least raise the issue so spork doesn't become too much of a mishmash of things.

pepe:
The first issue here is to look at Temple as HTML templates. It could be used for any templating. Even Janet code files, as could be seen here: https://git.sr.ht/~pepe/chidi/tree/master/item/chidi/generate. Or let's say benchmark results?

The second issue is the imagined spork library size, where we all differ a lot. But from some past conversation and by the latest confirmation https://github.com/janet-lang/janet/discussions/597#discussioncomment-296984 it should be big. Temple is one hundred forty lines of code in one file if we talk about the code size.

The third is the value of the code in spork as an example or educational piece, which for any loader related problems, Temple is a treasure. Pun intended.

Finally, Calvin's opinion is the most important here, and it was positive to the merge.

MikeBeller:
@pepe this all makes sense.  I just now reread the doc and noticed the mention of non-html usage.  I'm for it now.

andrewchambers:
it will become harder and harder to do versioning properly if any of the sub libraries need backwards incompatible changes, i am not sure i like the direction spork is headed.

pepe:
Versioning is indeed an issue, but mostly for the future ourselves, and I believe we will come with a smart solution.

BTW @andrewchambers, would you agree with adding some of your small libraries into spork? URI comes to mind first, if @bakpakin would agree.

andrewchambers:
Hmm, uri and also bigint are perhaps candidates, though they both require C code. (URI has a tiny bit of C for performance, that could maybe be removed, but i don't really want to)

pepe:
Hmm, I guess spork should stay .janet only, but only personal opinion.

andrewchambers:
perhaps if jpm could decide if you had a c compiler and fall back.

bakpakin:
I think it is rather useful. Also, spork libraries don't need to be completely generic in my opinion, and I think temple is unlikely to change much of at all.

PR #35, opened 2021-03-31
Add `temple/compile` function

subsetpark:
This exposes the same temple templating functionality as a function that takes a template string.

This allows on-the-fly templating without creating a template file.


bakpakin:
Looks pretty good, but should probably include the `where` parameter of `temple/create` as well.

subsetpark:
@bakpakin what do you think would be appropriate - the current file?

subsetpark:
@bakpakin added `dyn :current-file` to the params.

PR #39, opened 2021-05-26
Broke out docs from readme into indiv doc/*.mdz files

uvtc:
I think it will be much easier to make sense of the docs if each module in spork has its own doc page.

I don't know how to add the mendoza template, but this is a start to getting a spork mendoza docs site going.


uvtc:
Note, I didn't see any docs in the readme for the spork regex module, hence why there's no regex.mdz file present in this PR. I mostly cut/pasted portions of the readme into their corresponding new .mdz file, and then adjusted formatting from markdown to mendoza. Also changed heading levels so they'd start in the .mdz files as \<h2\>'s.

sogaiu:
FWIW, there is some info about `spork/regex` here: https://github.com/janet-lang/spork/blob/master/spork/regex.janet#L1-L23

uvtc:
Ok. Added a file containing that content. Put ``` @code`` ``` markers around the the regex code content.

bakpakin:
LGTM.

PR #44, opened 2021-08-07
Add streams module

subsetpark:


spork/generators.janet
```diff
@@ -0,0 +1,107 @@
...
+
+(defn cycle
+  "Repeatedly yield the elements of `ds`, looping back to the beginning when finished."
+  [ds]
+  (coro
+    (def to-cycle (if (indexed? ds) ds @[ds]))
+    (var i nil)
+    (while true
+      (set i (next to-cycle i))
+      (if (nil? i) (set i 0))
```
pepe, line 106: I guess `i` should be set to `(next ds nil)`:

```
repl:46:> (def h @{:A :B :C :S})
@{:A :B :C :S}
repl:47:> (next h nil)
:C
repl:48:> (next h :C)
:A
repl:49:> (next h :A)
nil
repl:50:> (next h 0)
:A
repl:51:> (next h :A)
nil
```

pepe (changes requested):
It looks good, but the cycle for the table issue.

spork/generators.janet
subsetpark: Good call 👌

pepe (approved):

bakpakin:
Looks like a good start. Definitely some more functionality could be added (for example, `map` could be variadic like in core) but we can add that later.

PR #45, opened 2021-10-04
Fix reading request from connection in http

pepe:
Namely, I removed the dst-end argument in pre-pop, as we want it to the end anyway. And I removed the pre-pop from read-body, as the buffer is already pre-poped.

Also, fix the default port and some typos.

This PR also makes some private functions public, as developers can use them in other libraries. If it is not desirable, please let me know, and I remove these changes.

### DIFF
--- a/spork/http.janet
+++ b/spork/http.janet
@@ -10,8 +10,7 @@
 (defn- pre-pop
   "Remove n bytes from front of buffer"
   [buf n]
-  (def m (- (length buf) n))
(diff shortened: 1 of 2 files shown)


bakpakin:
Might be good to get investigate why large requests fail, but it otherwise this looks good.

pepe:
From what I investigated, it is something in the net.c, but I got it on my to-do list. Thank you very much.

PR #46, opened 2021-10-06
Add concat chunk

pepe:
In my web development, I have found that many times I concat some pre-rendered strings from the helpers:

```
{- (string (link "somewhere") (do-not-click "somewhere link")) -}
```

So I thought it would be good to have a particular type of chunk that does just that:

```
{. (link "somewhere") (do-not-click "somewhere link") .}
```

What do you think?

### DIFF
--- a/spork/temple.janet
+++ b/spork/temple.janet
@@ -78,6 +78,12 @@
     (code-chunk
       (string "\n(prin (do " str "\n)) ")))
 
+  (defn concat-chunk
+    "Insert concat chunk into parser"
+    [str]
+    (code-chunk
+      (string "\n(prin (string " str "\n)) ")) )
+
   (defn string-chunk
     "Insert string chunk into parser"
     [str]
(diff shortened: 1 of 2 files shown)


andrewchambers:
I guess one question is how to concat an escaped string?

pepe:
I got `(def . string)`, yet I feel this is a better solution.

Ad escaping: I am using it only when gluing already escaped strings.

pepe:
After one more iteration, I refactored `print-chunk` and made all the `raw-` and `sub-` chunks use it.

I also added discrete tests for all the chunks, which could make this PR worth merging :-D

pepe:
NP, if nomerge :-)

bakpakin:
Yeah, I actually won't merge this, for the reason that the solution I proposed is actually more efficient and doesn't create the intermediate string, and therefore I want to encourage that usage instead.

PR #58, opened 2021-12-16
Add schema module to init for importing

pepe:
This PR adds `schema` module to `spork/init` for convenient import.

### DIFF
--- a/spork/init.janet
+++ b/spork/init.janet
@@ -10,5 +10,6 @@
 (import ./path :export true)
 (import ./regex :export true)
 (import ./rpc :export true)
+(import ./schema :export true)
 (import ./temple :export true)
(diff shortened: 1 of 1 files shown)


pepe:
I checked the failing test, and it is bizarre as it is in the temple test, which this PR is not touching. On my machine, the test is not failing.

sogaiu:
Similar results here (another Linux box) -- when I try the changes locally, `jpm test` shows "All tests passed.".

sogaiu:
Looking at the error output, may be it's this line here?

https://github.com/janet-lang/spork/blob/master/test/suite0008.janet#L16

pepe:
@sogaiu yep, that I guess is the offending line, but I have no idea why

PR #67, opened 2022-07-18
Add more docs on htmlgen

pepe:
As I read through the code of htmlgen, I added some more docs for public API, where I found it appropriate.

I also changed the call to `render` in the `html` function, as it seems more logically consistent.

And a small question: how do you intend to use `doctype-html`?

### DIFF
--- a/spork/htmlgen.janet
+++ b/spork/htmlgen.janet
@@ -107,7 +107,8 @@
 ###
 
 (defn raw
-  "Get an object that can be used to splice in HTML literals"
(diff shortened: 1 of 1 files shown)


bakpakin:
> And a small question: how do you intend to use doctype-html?

Intended use is to stick it at top of file. html-gen can take arrays to make a sequence of html fragments together, as opposed to a tuple which is a tag.

pepe:
That I overlooked, thank you very much.

PR #70, opened 2022-08-19
Add misc/randomize-array and sh/exec-slurp functions

tionis:
Just added these two simple functions as mentioned in the Gitter/Matrix chat.

### DIFF
--- a/spork/misc.janet
+++ b/spork/misc.janet
@@ -125,8 +125,8 @@
 
 (defn randomize-array
   "Randomizes array using the fisher-yates shuffle, takes an optional random number generator"
-  [arr &opt rng-in]
-  (def rng (if (= rng-in nil) (math/rng (os/cryptorand 8)) rng-in))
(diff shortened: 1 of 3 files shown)


spork/misc.janet
```diff
@@ -121,3 +121,17 @@
       (each child node (dfs child visit-leaf node-before node-after get-children seen))
       (when node-after (node-after node)))
     (visit-leaf data)))
+
+
+(defn randomize-array
+  "Randomizes array using the fisher-yates shuffle, takes an optional random number generator"
+  [arr &opt rng-in]
+  (def rng (if (= rng-in nil) (math/rng (os/cryptorand 8)) rng-in))
```
pepe, line 129: Here I would recommend`default` macro.

tionis: Ah yes, when I wrote this code, I didn't know about that.
Should be fixed now.

PR #72, opened 2022-08-24
Added new string module with trim-prefix and trim-suffix

tionis:
I wasn't sure about the module name, please tell me if it doesn't fit.


bakpakin:
Probably best to just put that in misc. No need for so many individual files.

tionis:
Ok, I moved it into misc

pepe (changes requested):
I think you should add tests for these additions.

tionis:
Fair point, I added two tests for each function

sogaiu:
Minor point -- the docstrings begin with lowercase letters.  May be it's slightly better to use "Trim" rather than "trim"?

tionis:
You're right, I fixed that

pepe (approved):
Thank you!

PR #73, opened 2022-08-31
Make rawterm example work with getch

pepe:
`rawterm/getch` returns buffer and not char.


bakpakin:
Thanks, forgot to update that one.

PR #74, opened 2022-08-31
getline: Add C-W for rubout-word, C-D delete when not EOL

paulsnar:
Fairly minor change, but I use these a lot so it would be appreciated if they were supported as well.


pepe (approved):
Another user here and LGTM

PR #76, opened 2022-09-04
getline: Improve Unicode support

paulsnar:
Okay, as mentioned in Gitter the other day, I have some improvements for getline that I want to share. This PR comprises enhanced support for Unicode and wide characters.

Some ideas about how to do this come from [Bestline](https://github.com/jart/bestline) (which I was in progress of attempting to port when getline was dropped into spork; and evolving getline is quite a bit easier than minifying bestline.) ~~The triple license notice in getline.c is due to the table being borrowed from the output of Cosmopolitan Libc, essentially copying its inner database -- please see 5defd38 for some context. That said, I'd like to rework the table generation tooling to depend directly on Unicode data, in which case that notice will no longer be necessary.~~ (This has been resolved.)

~~This is currently a draft because it still needs unit tests at least for utf8 and _getline, and possibly getline itself; I hope to get around to them in the coming couple of days.~~ (Tests for getline itself will come in a later PR.)

~~(Also the commit history is kind of noisy, it might be better to squash it.)  ~~
~~(Also I haven't tested all the intricacies yet with things like history and autocompletion with Unicode contents. Again, hence why it's a draft.)~~


paulsnar:
Short note: The idea I had on how to test getline was essentially to:
* Mock out rawterm via the module loader;
* Probably have rawterm/{begin,end} be noops;
* Have rawterm/getch return a predefined character instead (or maybe read from a channel or something, idk);
* Have rawterm/size return a predefined size;
* Set *err* to be a buffer;
* Assert that for a given sequence of inputs and size, a sensible output is produced.

It's the best I could think of without introducing needless layers of abstraction in getline whose only real purpose would be to enable testing.  If there are simpler ways of doing the same, please let me know.

The point is that it feels like getline should be tested holistically (or end-to-end, as the youths say) from what the user would see at the terminal, as well as the individual components where it makes sense.

bakpakin:
Looks good! Definitely some questions though.

At this point, what code is from Bestline? I literally don’t see any. The character tables seem to be auto generated.

Also I don't see why the print wrapper code with protect-term is really needed.

paulsnar:
> At this point, what code is from Bestline?

Fair cop—there's no code from Bestline (or Cosmopolitan, for that matter); the notice was targeted at an earlier version (pre 9ea225755f02f72841060349888362894105df7f) and by this point can probably be removed. (Given that IANAL, I'm not entirely sure what the status of the data itself is, regarding database rights, especially for Unicode data, so I'd leave a notice in there still, but the final call is yours.)

> I don't see why the print wrapper code with protect-term is really needed.

It's part of supporting what I called "async print", that is, allowing other fibers to print output while `getline` is still in progress without messing up the getline state or making it look weird. Doing this requires either calling these getline-prefixed functions or setting \*out\*/\*err\* in the fiber env though.

Now that I think about it, it would probably be better served as a separate PR (if it even has a place in getline at all), the fact that it made its way into this one is essentially an accident on my part, so I'll remove it from this PR.

paulsnar:
(Something appears to have [broken](https://github.com/janet-lang/spork/runs/8186986633?check_suite_focus=true) with CI and native modules. Same error can be observed on master of this repo, and I can't reproduce locally.)

paulsnar:
Okay, sorry, but I'm postponing tests for getline proper to a separate PR as well. @bakpakin Please take a look again, thank you.

bakpakin:
Issue was a broken monkey patch in janet-lang/jpm, which has been reverted.

bakpakin:
LGTM! Thanks for all of the work on this, especially the testing code! Looks like the code is ready to me merged and developed on the main branch.

PR #79, opened 2022-09-23
added spork/relpath

tionis:
Returns the relative path between two subpaths
this should imitate the behaviour of e.g.
```bash
realpath --relative-to="a/bit/deeper/with/some/nested/dirs" "a/bit/deeper/with/other/nested/dirs"
```

I wrote a few tests for it in a new test suite.

### DIFF
--- a/spork/path.janet
+++ b/spork/path.janet
@@ -25,6 +25,14 @@
   [from to]
   (setdyn (symbol to) (dyn (symbol from))))
 
+(defn- parent? [source-parts target-parts]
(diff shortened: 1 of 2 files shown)


bakpakin:
LGTM

pepe:
@tionis, it would be great if you added some documentation to `content/path.mdz` for this new functionality.

tionis:
Ok, I added the documentation in #80 and also fixed a bug I discovered in my implementation.

PR #82, opened 2022-09-29
Add assert-docs test and fix, what it found in spork/test

pepe:
As promised, I am starting to move codes from my libraries to the spork. I am starting small by adding one function to the `spork/test`. `assert-docs` checks that all the public bindings from the environment of the required path have a docstring. I have already fixed many of the docstrings in my libraries and the `spork/test` in this commit. 

I am leaving this as a draft cause I would like to check all the modules in the spork as a part of this PR if you find it worthy.

### DIFF
--- a/doc/test.mdz
+++ b/doc/test.mdz
@@ -139,3 +139,14 @@ Runs the form, but supresses its stderr.
 (suppress-stderr (eprint "Hello world!"))
 # => nil
 ```
+
+## assert-docs
+
+Asserts that all public bindings from the environment have docstring, 
+when the `path` is required.
+
+@codeblock[janet]```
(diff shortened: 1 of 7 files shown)


pepe:
I already went ahead and asserted docs in all modules. I am making it ready for review.

bakpakin:
LGTM. Would be nice if we could have some way to signal to format that a macro should be treated as a block macro rather than a function call, but not that important.

pepe:
Thank you! May you please elaborate a little about the block macro vs. function call? I m not sure what you mean here. Anyway, I commence with the subsequent additions, which will be more substantial.

pepe:
But I do not see any simple way with the current implementation to fix this.

PR #83, opened 2022-09-30
Move misc from marble, rename and deprecate

pepe:
In the pursuit of moving my codes to spork, I have added many to the misc module. It ranges from simple mapping functions to PGP word generators.

The peg-grammar is the special addition I would like to pick for better investigation by the reviewer. I meant it as a grab bag of the peg rules and a simple mechanism to use them in your code. Presumably, we will add much more to this registry.

I also streamlined docstrings in the whole module with a simple rule: if it fits on one line, use " if it is not, use ``` on its line.

I also added prefixes to a couple of functions already in the module (with the old name as an alias, with a deprecation flag), which is out of scope for this commit but plays well with the new additions.

All the new functionality is tested and documented.

### DIFF
--- a/doc/misc.mdz
+++ b/doc/misc.mdz
@@ -60,3 +60,325 @@ x       y
   banana  orange
      cat     dog
 ```
+
+## map-keys
+
+Returns new table with function applied to table's keys recursively.
+
+@codeblock[janet]```
+(misc/map-keys string {1 2 3 4}) 
+# => @{"1" 2 "3" 4}
+```
+
+## map-vals
+
+Returns new table with function applied to table's values.
+
+@codeblock[janet]```
+(misc/map-vals string {1 2 3 4}) 
+# => @{1 "2" 3 "4"}
+```
+
+## select-keys
+
+Returns new table with selected keys from dictionary.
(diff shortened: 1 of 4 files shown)


doc/misc.mdz
```diff
@@ -60,3 +60,325 @@ x       y
...
+
+## caperr
+
+Captures the standart error output of the variadic body and returns it as
+a buffer.
+
+@codeblock[janet]```
+(misc/capout
+  (def p "HOHOHO")
+  (prin p))
```
l63l, line 176: needs edit after copy&paste

pepe: I have looked at this part and do not see any problem. May you please be more specific? I found another problem with a block on line 187 though :-D.

l63l: it is the example from capout copy&pasted to caperr and not:
```
(misc/caperr
  (def p "HOHOHO")
  (eprin p))
```

pepe: Oh, the programmer's blindness. Thank you very much!

bakpakin:
LGTM

PR #85, opened 2022-10-02
added base64 encoder and decoder

tionis:
I added it in its own module, please tell me when I should move it somewhere else in the namespace.  
I also added tests for it.

### DIFF
--- a/doc/misc.mdz
+++ b/doc/misc.mdz
@@ -384,3 +384,19 @@ Returns an array of hexs from the string with pgp words.
 (misc/pgp/words->hexs "absurd-scavenger accrue_universe upshot.village")
 # => @["01" "D1" "02" "EE" "F4" "F6"]
 ```
+
+## base64/encode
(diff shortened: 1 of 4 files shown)


pepe:
I had a similar feeling for the PGP words, so I would say misc is ok.

pepe (changes requested):
Please add `assert-docs` to your tests. You will find something I guess. Also, I do not see the reason for docstrings for functions, which are private.

tionis:
I moved the code into misc, fixed the docstrings and added docs

tionis:
I also merged my tests into suite 20 so the tests for misc are grouped together

bakpakin:
Why is this in misc when it could just be in a library "spork/base64"?

I agree with an early observation that spork is already a misc library, so the naming is redundant

tionis:
OK, I moved the code into base64.janet again and also migrated the tests into a new test suite

PR #92, opened 2022-11-06
Added copy and devnull from jpm/shutil to spork/sh

tionis:
These three functions seem useful and were included in jpm's shell helper.
I had integrated spork/path into spork/sh/copy but otherwise the functions are nearly unchanged

### DIFF
--- a/spork/sh.janet
+++ b/spork/sh.janet
@@ -5,6 +5,11 @@
 
 (import ./path)
 
+(defn devnull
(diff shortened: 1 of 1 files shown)


bakpakin:
LGTM. The `copy` function was always a hack to me, but it does work

tionis:
I also don't really like it, but it's been working quite well in jpm it seems and it is convenient not to have to read everything into memory etc.

Issue #98, opened 2023-01-01
sh/make-new-file (mostly) ignores mode argument


sogaiu:
In https://github.com/janet-lang/spork/commit/ae9db3d5c953d54a0cca74f9352db160e749fce9, it looks like some steps were made toward making it possible for a caller of `sh/make-new-file` to specify a mode when it opens a file.

However, the mode argument appears to be unused by the `file/open` call: https://github.com/janet-lang/spork/blob/63f656e6f41b469b484903373187e96616eac837/spork/sh.janet#L95-L104

sogaiu:
#99 was merged so I'm closing this.

PR #99, opened 2023-01-01
Use mode argument in make-new-file

sogaiu:
An attempt to address #98.

### DIFF
--- a/spork/sh.janet
+++ b/spork/sh.janet
@@ -101,7 +101,7 @@
     (when (and (not (exists? file-path))
                (not (exists? parent-path)))
       (create-dirs parent-path)))
-  (file/open file-path :w))
+  (file/open file-path mode))
 
 (defn copy-file
   "Copy a file from source to destination. Creates all directories in the path


tionis:
I'm not sure, mode is means the file open flags, mode normally refers to the permission bits, as far as I recall

sogaiu:
I'm just going by the code that was mostly there already (I think it got put there in https://github.com/janet-lang/spork/commit/ae9db3d5c953d54a0cca74f9352db160e749fce9#diff-addba8464816d2a707000ee1f01e77510dc484a709013f5b62979b3834ef8f11R76) as well as the docs:
```
    (file/open path &opt mode)

    Open a file. `path` is an absolute or relative path, and `mode` is 
    a set of flags indicating the mode to open the file in. `mode` is a 
    keyword where each character represents a flag.
```
Or may be I'm missing something?

tionis:
Oh, I stand corrected, at least in the context of Janet file handling, mode seems to mean exactly what you suggest.

pepe (approved):
LGTM

PR #103, opened 2023-01-17
Add tabseq to indent-2-forms

sogaiu:
Since `loop` and `seq` are in `indent-2-forms`, perhaps their friend `tabseq` could be too?


pepe (approved):
LGTM

PR #114, opened 2023-05-09
:pointer case didnt work:

kamisori:
it printed "<tuple 0x023B4A2F7E40>" instead of "void*" with rtype being tuple: (* void)

now it prints type* in that case, so void*
or JanetBuffer* in case of :buffer ...
or const char* in case of cstring


bakpakin:
Wouldn't `emit-type` be the better choice here rather that this hack for pointers only?

kamisori:
please see #118

PR #116, opened 2023-05-09
i is incremented on encountering any of & &opt

kamisori:
&key etc. but length of paramlist only on found params

without this, optional parameter get-calls use wrong index:

```
 (c/cfunction create-window
               :static
     ""
     [width:int
      height:int
      title:cstring
      &opt
      (monitor (* GLFWmonitor) NULL)
      (share (* GLFWwindow) NULL)] -> :pointer
```
gives:
```
...
const char * title = janet_getcstring(argv, 2);
void *monitor = janet_optabstract(argv, argc, 4, GLFWmonitor, NULL);
...
```

expected a 3 in that last line

### DIFF
--- a/spork/cjanet.janet
+++ b/spork/cjanet.janet
@@ -727,8 +727,8 @@
         (array/push
           argument-parsing
           (if found-optional
-            (janet-opt* p 'argv 'argc i param-names cparams)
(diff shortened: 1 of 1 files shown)


bakpakin:
Hmm, good catch. This code does need some more testing, but otherwise LGTM.

PR #120, opened 2023-05-16
fixed bug in sh/copy and added sh/exec and sh/exec-fail

tionis:
If we have exec-slurp it makes sense to also have a simple exec in sh in my book.
Also fixed a bug in sh/copy


sogaiu:
Don't have an opinion on the new functions, but the fix looks good :)

PR #121, opened 2023-05-18
Fixed 4-byte utf-8 encoding

zevv:


sogaiu:
Would using U+1F44D instead for the test work?

bakpakin:
LGTM either way - I will be more responsive on the weekend.

PR #122, opened 2023-05-23
Added shell like string splitting to spork/sh

tionis:
I imported the code from andrewchamber's [janet-shlex](https://github.com/andrewchambers/janet-shlex) (with permission).
It allows splitting a string following shell quoting rules.


tionis:
Fair point, I changed it to `escape`

PR #123, opened 2023-05-26
for loop was borked

kamisori:
it printed the body doubly when it wasnt empty


kamisori:
see this for an usage example:

https://github.com/kamisori/wjpu/blob/fae40b6cdc45be4312eb84f00b90c1f651f720fb/src/glfw.janet#L189

bakpakin:
Wow that’s pretty cool. Fix is sound - the if empty logic looks a bit unneeded but that’s more of a “clean code” issue.

PR #124, opened 2023-05-30
Fix a small bug in math/permutations.

otrho:
The `math/permutations` implementation uses a non-recursive Heap's algorithm and had a typo checking if `k` was even instead of `i`, producing slightly incorrect results.


pepe (approved):
LGTM. My bad, thanks for catching it.

PR #133, opened 2023-06-26
tests failed due to windowsy newlines in literals, but unixy in output

kamisori:
on windows: tests failed due to windowsy newlines in literals, but unixy in output


sogaiu:
How about `eol` as a name over `newline`?

sogaiu:
Thanks for the change!

PR #142, opened 2023-07-13
idk why but i couldnt rebase

kamisori:


kamisori:
see https://github.com/janet-lang/spork/pull/141

kamisori:
added a fix for spork/math/quantile to make it play along with typed arrays + tests

PR #143, opened 2023-07-18
add binary-search macros

primo-ppcg:
- Adds `binary-search` and `binary-search-by` to efficiently locate a value / insert position within a sorted tuple or array. These are implemented as macros so that primitive comparators will compile directly to vm ops.
- Updates `insert-sorted` and `insert-sorted-by` to utilize these macros.
- `misc.janet` has been run through `fmt/format-file`.

No new tests have been added, the current tests for `insert-sorted` and `insert-sorted-by` are likely sufficient.

### DIFF
--- a/spork/misc.janet
+++ b/spork/misc.janet
@@ -88,7 +88,7 @@
   # build horizontal bars
   (def topbuf (or buf-into @""))
   (buffer/push topbuf "╭")
-  (def hbuf   @"│")
+  (def hbuf @"│")
   (def midbuf @"╞")
   (def botbuf @"╰")
   (each key colkeys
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

sogaiu:
I tried it out a bit.  In my limited testing it seemed fine :+1: 

I wondered a bit about the case where one might want to know if some element is present.

primo-ppcg:
> one could inspect the array using the returned index

Seems sufficient to me, although a separate implementation could implement short circuiting. Worst and average runtime complexity would remain unchanged, though.

PR #144, opened 2023-07-20
Added abstract type set

tionis:
Based on work done by @ianthehenry in [janet.guide](https://janet.guide/xenofunctions/) (which he said he considers public domain).
This implements a simple mutable set based on the JanetTable data structure.
This is mainly a draft to ask for opinions on differentiating between mutable/immutable sets (like done for other data structures in janet) and also a few of the pain points listed below:
- hashing is currently done by casting to struct, is there a better way?
- comparisons are currently done by hash, which is probably not what a user would expect? (it does allow checking for equality of two sets though, due to the deterministic hashing)
- I'm not sure if the cloning is optimal (can a memcpy be user here for more efficient cloning?)

### DIFF
--- a/project.janet
+++ b/project.janet
@@ -40,6 +40,10 @@
   :source @["src/utf8.c"])
 
 (declare-native
+  :name "spork/set"
+  :source @["src/set.c"])
+
+(declare-native
  :name "spork/tarray"
  :headers @["src/tarray.h"]
  :source @["src/tarray.c"])
--- a/src/set.c
+++ b/src/set.c
@@ -1,0 +1,318 @@
+#include <janet.h>
+
+static int set_gc(void *data, size_t len) {
+  (void) len;
+  janet_table_deinit((JanetTable *)data);
+  return 0;
+}
+
+static int set_gcmark(void *data, size_t len) {
+  (void) len;
+  janet_mark(janet_wrap_table((JanetTable *)data));
+  return 0;
+}
+
+static void set_tostring(void *data, JanetBuffer *buffer) {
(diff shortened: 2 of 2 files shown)


ianthehenry:
I would expect for `compare` to only compare the pointers of the data types, since it's a mutable structure -- that would be consistent with the behavior for tables and buffers and arrays.

We could simplify this a little bit by adding some private `wrap` and `unwrap` functions on the abstract type that return the underlying table, and then implement functions like `symmetric-difference` and friends in Janet code instead of C. (I have nothing against C).

Also worth noting that this does rely on sort of a private implementation detail of the Janet GC. We could fix that, at the cost of an extra pointer indirection. I don't have a good sense for how fragile it is.

tionis:
Just a small note: As soon as I find the time I will split up the implementation into an immutable and mutable version mirroring structs and tables regarding the hashing behavior.
(I will also add a function to convert between the two)
I'm unsure on how to label the namespace, probably something like `set/mutable/new` and `set/immutable/new` etc., but I'm not really sure yet. This naming scheme seems too verbose and inelegant.

bakpakin:
Not sure this is really something that useful as it is no more efficient than a Table, a good example for how to make abstract types but not something I would actually use.

Also some questionable behavior with `set_get` - it looks as though you can't put numbers in the set?

tionis:
Valid points.
Also important is that the additions of helper functions like map-keys did reduce the benefit of this quite a bit. 
I also don't have time at the moment to complete this, so I'll close this.

Issue #140, opened 2023-07-12
Missing prime number generator


sogaiu:
Is there any interest in including a prime number generator?

I don't know about generating [this one](https://codegolf.stackexchange.com/questions/146017/output-the-ppcg-prime), but I learned recently that some hash function constants are computed from some primes so may be there are other use cases too?

I wonder if @primo-ppcg's [gist with a fiber-based one](https://gist.github.com/primo-ppcg/31a31ecede0296457b6c57fe1f0c8020) would be a candidate...

primo-ppcg:
I actually intended to add three primality functions to `spork/math`:

`(math/primes)` a boundless prime number generator
`(math/prime? n)` a primality test, deterministic for n < 2^64
`(math/next-prime n)` next prime strictly larger than n

I was unable to implement the last two, for want of a modular exponentiation which avoids overflow, which I intended to implement first. I suppose it wouldn't hurt to add `math/primes` on its own, though.

primo-ppcg:
Work in progress: https://gist.github.com/primo-ppcg/31a31ecede0296457b6c57fe1f0c8020

There's a few things I'd like to clean up yet, `lucas-prp?` would be much cleaner implemented recursively, and I'd really like to have `mulmod` and `powmod` in a C module, as they're used pretty much everywhere.

It'd also be great to have a good home for this:
```janet
(defn binary-search
  ``Returns the index of `x` in a sorted array or tuple
  or the index of the next larger item if `x` is not present.
  This is the correct insert index for `x` within `arr`.``
  [x arr]
  (var start 0)
  (var end (length arr))
  (var mid (brshift end 1))
  (while (not= mid end)
    (if (< (arr mid) x)
      (do (set start mid) (set mid (brshift (+ start end 1) 1)))
      (do (set end mid) (set mid (brshift (+ start end) 1)))))
  mid)
```
I need it for both `prime?` and `next-prime`, but it doesn't really belong in a math module.

sogaiu:
Re: `binary-search`

I know of a few things in similar territory:

* https://github.com/Duncaen/adventofcode-janet/blob/fff54095eaa5acb47669b9b0e124c3e8f61f4ff6/2020/util.janet#L11-L24
* https://github.com/saikyun/freja/blob/main/freja/find_row_etc.janet
* https://github.com/saikyun/janet-binary-search-comparison

So evidence of a common need, perhaps?

Not sure where it would go, but may be others have some ideas about that.

---

[...]

PR #145, opened 2023-07-24
Add functions for primality testing

primo-ppcg:
Related: #140

Adds the following functions for primality testing:
- `(math/prime? n)`
A primality test, deterministic for all `n` less than 2^63.
- `(math/next-prime n)`
Returns the next prime number strictly larger than `n`.
- `(math/primes)`
A boundless prime number generator.

Also defines the following support functions in a separate C module:
- `(math/powmod a b m)`
Modular exponentiation of `a` to the power of `b` mod `m`.
- `(math/mulmod a b m)`
Modular multiplication of `a` and `b` mod `m`.
- `(math/invmod a m)`
Modular multiplicative inverse of `a` mod `m`.
- `(math/jacobi a m)`
Computes the Jacobi Symbol (a|m).

~~I haven't added any tests to the PR yet~~ (now added), but it has passed this battery of tests: [prime-tests.janet](https://gist.github.com/primo-ppcg/2e03d79de4583be0f0f555e0b5cf0e9a). This could be stripped down some before including in a test suite.

Prime testing for primes on the range 2^53 - 2^63 is about 16µs per prime on my system. Composites on average are much faster. This is down from around 600µs using a pure Janet implementation.

### Possible Concerns
The C module is imported without prefix into the `spork/math` module. I don't know if this is the proper way to do it.

128-bit operations are used for modular multiplication where available. On Linux, I test with:
```c
#if defined(__SIZEOF_INT128__)
```
which is also used in the [Linux kernel](https://github.com/torvalds/linux/blob/6eaae198076080886b9e7d57f4ae06fa782f90ef/include/linux/math64.h#L161C53-L161C70), so it's probably sufficient.

For Windows I test with:
```c
#elif defined(_WIN64) && defined(_MSC_VER)
```
which I can confirm works with VC 2019 and VC 2022.

The C functions use signed 64-bit integers in order to allow potentially negative arguments, which unfortunately means these functions (and therefore also the Janet functions that rely on them) don't work properly with `int/u64` values exceeding 2^63. I think this could possibly be addressed, but not without a lot of code duplication in the C module.


sogaiu:
> it has passed this battery of tests: [prime-tests.janet](https://gist.github.com/primo-ppcg/2e03d79de4583be0f0f555e0b5cf0e9a).

I gave these a try here (manually as well), and unsurprisingly, they passed :)

Also tried some random-ish invocations successfully.

My attempts were on a Linux machine.  I don't have access to macos or any BSD machine -- I wonder if we can get someone to test on some of those.

> Prime testing for primes on the range 2^53 - 2^63 is about 16µs per prime on my system. Composites on average are much faster. This is down from around 600µs using a pure Janet implementation.

Nice!

> The C module is imported without prefix into the spork/math module.

That seems ok to me.

> The C functions use signed 64-bit integers in order to allow potentially negative arguments, which unfortunately means these functions (and therefore also the Janet functions that rely on them) don't work properly with int/u64 values exceeding 2^63. I think this could possibly be addressed, but not without a lot of code duplication in the C module.

Perhaps this can be left as an exercise for the interested user down the line?

sogaiu:
I got NetBSD running via qemu and tried out the PR and it seemed to worked fine.

Building was ok and the `prime-tests.janet` file gave no errors :+1:

sogaiu:
FWIW, ran the tests for [e74b793](https://github.com/janet-lang/spork/pull/145/commits/e74b7934328bd25c8c938c3cf1ea47b935a3d648) on Linux and NetBSD and things seemed fine :+1:

Techcable:
How common is testing for primes? What use case are you thinking of that makes it deserve to be in spork?

(Just curious, not trying to be combative...)

primo-ppcg:
> How common is testing for primes? What use case are you thinking of that makes it deserve to be in spork?

The most useful function is likely `next-prime`. A number of applications require one or more primes, often of a similar length. This of course also relies on `prime?`. The `miller-rabin-prp?` is perhaps a bit too niche, and wouldn't need to be exported necessarily.

sogaiu:
It was mentioned in #140 -- that queried about interest in prime number generation -- but some hash functions have initial values that depend on primes.

I ran into [this](https://github.com/sogaiu/janet-checksums/blob/b0d18fb1a0719c629bae2d27d14420cf293f9817/janet-checksums/sha-2.janet#L11-L30) when implementing some of the SHA-2 functions.

Granted, that's not testing for primality (as asked about above), but rather generating primes (^^;

primo-ppcg:
> It was mentioned in #140 (where there was a query about interest in prime number generation) that> I ran into [this](https://github.com/sogaiu/janet-checksums/blob/b0d18fb1a0719c629bae2d27d14420cf293f9817/janet-checksums/sha-2.janet#L11-L30) when implementing some of the SHA-2 functions.

```janet
(take 64 (math/primes))
```
would definitely look cleaner, although with so few, listing them out is fine too.

sogaiu:
> (take 64 (math/primes))

> would definitely look cleaner, although with so few, listing them 
> out is fine too.

Yes, I agree.

Using a form like the aforementioned allows one to skip the step of manually enumerating as well as verifying correctness of individual numbers and whether any had been skipped or duplicated (^^;

sogaiu:
Tests from [b828418](https://github.com/janet-lang/spork/pull/145/commits/b828418e642d609ff0b765f9cb7d5c17a8e0653d) ran with no issues here (Linux and NetBSD) :+1:

primo-ppcg:
> Tests from [b828418](https://github.com/janet-lang/spork/pull/145/commits/b828418e642d609ff0b765f9cb7d5c17a8e0653d) ran with no issues here (Linux and NetBSD) +1

Appreciated. I apologize for the force pushes, I just want to avoid a bunch of minor commits in the commit history, should this PR be merged.

sogaiu:
I noticed today that janet itself has [a primes example](https://github.com/janet-lang/janet/blob/a13aeaf9557250639018802baceeefb78b834846/examples/primes.janet).  Just mentioning it as there having been some additional interest in the past :)

PR #148, opened 2023-08-11
Update `generators/to-array`

primo-ppcg:
Update `generators/to-array` to consume its argument into an array, as documented.


sogaiu:
Works here :+1:

PR #149, opened 2023-08-14
Add `merge-sorted`, `merge-sorted-by`

primo-ppcg:
Adds `merged-sorted` and `merge-sorted-by` to merge two arrays which are known to be sorted previously.

This is more convenient, and also faster than either:
- concatenating the arrays, and applying `sort` to the result
- splicing the second array into `insert-sorted`


pepe (approved):
LGTM

primo-ppcg:
A side note: the parameter order for `merge-sorted-by` is consistent with `insert-sorted-by`. However, both are _inconsistent_ with the core function `sorted-by` (or every `-by` function for that matter, the first parameter is always `f`).

It's probably better for parameter order to be consistent within a module, but perhaps both should be updated.

PR #153, opened 2023-09-01
Code quality improvements, bugfixes, additional tests for `spork/data`

CFiggers:
Added additional tests to /test/suite0021.janet , applying to `spork/data/diff`.

This revealed bugs in the way `diff` was handling larger, more deeply-nested structures.

While fixing bugs, simplified `diff`'s logic which allowed eliminating several small "helper" functions. The eliminated function calls *probably* improve performance of `diff`, but I haven't benchmarked before/after.

All tests, including newly-added ones, now pass.

### DIFF
--- a/spork/data.janet
+++ b/spork/data.janet
@@ -34,24 +34,22 @@
   (let [va (safe-in a k)
         vb (safe-in b k)
         [a* b* ab] (diff va vb)
-        in-a (in? k (keys a))
-        in-b (in? k (keys b))
-        same (and in-a in-b
+        same (and (in? k (keys a)) (in? k (keys b))
                   (or (not (nil? ab))
                       (and (nil? va) (nil? vb))))]
-    [(when (and in-a (or (not (nil? a*)) (not same))) {k a*})
-     (when (and in-b (or (not (nil? b*)) (not same))) {k b*})
+    [(when (and (in? k (keys a)) (or (not (nil? a*)) (not same))) {k a*})
+     (when (and (in? k (keys b)) (or (not (nil? b*)) (not same))) {k b*})
      (when same {k ab})]))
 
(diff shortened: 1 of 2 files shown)


sogaiu:
May be minor point but I think [`start-suite`'s argument is currently optional](https://github.com/janet-lang/spork/blob/e48eaf451bd5ab76f3c7dcf1f11af7e85b0f0fb3/spork/test.janet#L44-L51).  May be it's not necessary to add one in this case?

I tried running the tests with and without an argument and the output didn't seem different wrt the presence of this argument -- though I might have missed something (^^;

When reorganizing janet's tests, [we made a similar change](https://github.com/janet-lang/janet/commit/9675411f352be882b575c642ef25202489e2461f#diff-1aa6b06257b8ade54827899f973f6cbb623a5bb2d444a2dc81aae2e75b527e37L37-R47) and I think it made it a bit easier for maintenance.  Possibly a similar thing holds here.

CFiggers:
I was just making this one match all the others that are currently out there. I noticed it because the output from `jpm test` is inconsistent between all the other test suites and suite0021:

image

I don't really have an opinion either way on which way is preferable.  ¯\_(ツ)_/¯

CFiggers:
...Of course, _now_ I notice that there's already tests out there organized by module instead of just being opaquely numbered. 😜

sogaiu:
>  I noticed it because the output from jpm test is inconsistent between all the other test suites and suite0021

Ah right -- I guess I did miss the difference :)

> there's already tests out there organized by module instead of just being opaquely numbered. 😜

I don't know if it would be seen as worth it, but I suppose we might be able to reorganize in a manner somewhat similar to what we did for Janet itself.

Though may be things are already gradually heading in that direction and it's another thing I failed to notice (^^;

PR #158, opened 2023-10-26
allow cfunctions in argparse :map

### LINKED ISSUE #157: in argparse, a cfunction cannot be used for :map
```
(use spork/argparse)
(pp (argparse nil "foo" {:kind :option :map |(scan-number $)})) # returns 3
(pp (argparse nil "foo" {:kind :option :map scan-number})) # returns "3"

```

tionis:
Fixes #157

### DIFF
--- a/spork/argparse.janet
+++ b/spork/argparse.janet
@@ -124,7 +124,8 @@
   (defn handle-map [name handler]
     (when-let [value (get res name)
                map-func (handler :map)
-               is-func (function? map-func)]
+               is-func (or (function? map-func)
+                           (cfunction? map-func))]
       (put res name (map-func value))))
 
   # Handle an option
--- a/spork/argparse.janet
+++ b/spork/argparse.janet
@@ -157,7 +157,8 @@
     (when-let [action (handler :action)]
       (cond
         (= action :help) (usage)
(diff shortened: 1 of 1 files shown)


sogaiu:
Thanks for this.

On a similar note, I wonder about [this bit](https://github.com/janet-lang/spork/blob/641b27238e073c5f5f963ec16c79b51643d3e66f/spork/argparse.janet#L159).  I'm not too familiar with argparse (^^;

Any thoughts?

tionis:
You mean :action?
This just executes the function at parse time

sogaiu:
I guess so.

The thing is, it's also just checking via only `function?` -- wondering whether it would make sense to change that too.

pepe (approved):
LGTM

pepe:
> The thing is, it's also just checking via only `function?` -- wondering whether it would make sense to change that too.

This makes sense. Another PR will be fine I guess. Or @tionis do you think you can add it to this one, please?

tionis:
> I guess so.
> 
> The thing is, it's also just checking via only `function?` -- wondering whether it would make sense to change that too.

You're right! I just added the change

pepe (approved):
LGTM

PR #166, opened 2023-12-29
create a table instead of an array to hold the fibers in pmap-full

mraveloarinjaka:
There is a crash when using pmap-full because the list of fibers passed to "join" is an array and not a table.
The fix aligns pmap-full implementation to pcall.
```
error: bad slot #0, expected table, got <array 0x600002C65C40>
  in table/clear [src/core/table.c] on line 411
  in drain-fibers [/Users/mraveloarinjaka/projects/sandbox/aoc2023/jpm_tree/lib/spork/ev-utils.janet] on line 35, column 3
  in join [/Users/mraveloarinjaka/projects/sandbox/aoc2023/jpm_tree/lib/spork/ev-utils.janet] on line 56, column 10
  in pmap-full [/Users/mraveloarinjaka/projects/sandbox/aoc2023/jpm_tree/lib/spork/ev-utils.janet] (tailcall) on line 84, column 3
```

### DIFF
--- a/spork/ev-utils.janet
+++ b/spork/ev-utils.janet
@@ -82,8 +82,9 @@
   (def chan (ev/chan))
   (def res (if (dictionary? data) @{} @[]))
   (join chan
-        (seq [[i x] :pairs data]
-          (ev/go (fiber/new (fn [] (put res i (f x))) :tp) nil chan)))
+        (tabseq [[i x] :pairs data
+                 :let [fib (ev/go (fiber/new (fn [] (put res i (f x))) :tp) nil chan)]]
+          fib fib))
   res)
 
 (defn pmap-limited


sogaiu:
That does look like a mismatch.

Using `tabseq` seems better and it's [what `pcall` (which also uses `join`) does](https://github.com/janet-lang/spork/blob/d644da0fd05612a2d5a3c97277bf7b9bb96dcf6b/spork/ev-utils.janet#L65-L75) -- perhaps that's something that motivated the specific proposed changes :)

Just spelling things out here for other reviewers.

mraveloarinjaka:
Hello,
What are the next steps?
It is not clear whether I should amend the changes or if I am missing something.

sogaiu:
My earlier comment was just saying that the proposed changes looked on the ok side.  Sorry if that was unclear.

I think this PR and some others are awaiting the main maintainer's availability -- AFAIU he is pretty busy.

PR #171, opened 2024-01-19
Feature: Add a Gaussian sampler

hyiltiz:
This adds a Gaussian sampler that exposes a similar interface to `spork/rand-uniform`, that is:
- can be called without args `(rand-gaussian)` for a single standard Gaussian sample
- can be called with args `(rand-gaussian 5 0.1)` for a single Gaussian sample with specified parameters
- can be used to generate `N` samples with the `sample-n` function, using `(sample-n |(rand-gaussian 5 0.1) 100)`
- does not leave any state pending (no fiber used) at the expense of wasting half the entropy exracted from the RNG device (does janet runtime rely on OS RNG entropy pool?)

### DIFF
--- a/spork/randgen.janet
+++ b/spork/randgen.janet
@@ -37,9 +37,7 @@
 
 (defn rand-gaussian
   "Get a random sample from the standard Gaussian distribution. 
-  Optionall specify the mean m and the standard deviation sd. E.g.:
-  (rand-gaussian) # => 0.1324...
-  (rand-gaussian 5 0.1) # => 5.3397...
+  Optionall specify the mean m and the standard deviation sd. 
   "
   [&opt m sd]
   (default m 0)
@@ -52,8 +50,10 @@
   # We use the Box-Muller transform
   (let [rho (math/sqrt (* -2 (math/log q)))
         theta (* 2 math/pi p)
-        # _box (* rho (math/cos theta))
         _muller (* rho (math/sin theta))
+        # in devices where hardware entropy pool usage should be efficient
+        # we can achieve x2 efficiency by using the `box` variable as well
+        # _box (* rho (math/cos theta))
         # box (scale _box)
         muller (scale _muller)]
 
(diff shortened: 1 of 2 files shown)


sogaiu:
I've been informed that this PR is sort of a "reboot" of [this earlier one](https://github.com/janet-lang/spork/pull/169) due to technical issues :)

sogaiu:
Seems to be working here -- at least the tests pass :)

Some minor points:

* Most (all?) of the existing docstrings don't contain example invocations.  I think sample invocations are helpful, but placing such things in tests is an ok alternative for spork.  I'd go for removing the examples from the docstrings of `rand-gaussian` and `sample-n`.  The current PR already contains some tests that reflect some of the examples from the docstrings so perhaps additional ones aren't necessary.  I'm not opposed though :)

* There appear to be some commented code bits remaining:
    ```janet
    (let [rho (math/sqrt (* -2 (math/log q)))
          theta (* 2 math/pi p)
          # _box (* rho (math/cos theta))
          _muller (* rho (math/sin theta))
          # box (scale _box)
          muller (scale _muller)]

      # (yield box)
      muller))
    ```
    Not sure what others will say, but if these are to stay, may be it makes sense to have some commentary explaining their presence.  I can see the value of mentioning that sort of half of what one might have expected is being discarded for future readers / maintainers [1].  Not a fan of the `yield` form even commented here.

---

[1] Wasn't clear on the choice to retain `muller` over `box` -- I think a previous version went with the reverse choice.

hyiltiz:
Updated w.r.t. feedback.

sogaiu:
I'm a bit fuzzy on this, but did we not conclude that using `yield` would lead to undesirable places wrt the function being usable in certain types of scenarios?  

If so, I think leaving [this comment](https://github.com/janet-lang/spork/pull/171/files#diff-5cc0ef9847d31c6c3a6a5c7ec3127e77efe6a1100bf5c942f50861c9223d90fbR60) is on the misleading side and potentially asking for trouble down the line.

sogaiu:
Thanks for the update.

LGTM.

hyiltiz:
Ready to merge and close it out.

Issue #173, opened 2024-02-02
Do `exec-slurp` and `exec-slurp-all` benefit from tighter resource control?


amano-kenji:
I wrote the following code because my janet fibers started crashing due to lack of pipes and files. I think garbage collector doesn't close pipes fast enough.

If pipes aren't closed, a long-running process will run out of pipes and files.

If I run the functions below in `ev/with-deadline` to prevent zombie processes from blocking janet, I still need to kill the zombie processes with `os/proc-kill`. Otherwise, I see zombie processes abandoned by janet.

```janet
(defn slurp
  ``
  It executes args with `os/spawn` and throws an error if the process returns with non-zero exit code.

  If the process exits with zero exit code, this functions returns standard output of the process.

  If there is any error, to prevent zombie processes, the spawned process is killed.

  Regardless of errors, the spawned process is going to be closed for resource control.
  ``
  [& args]
  (with [proc (os/spawn args :xp {:out :pipe})]
    (try
      (let [[out] (ev/gather
                    (ev/read (proc :out) :all)
                    (os/proc-wait proc))]
        out)
      ([err f]
        (try
          (os/proc-kill proc)
          ([_]))
        (propagate err f)))))

(defn slurp-all
  ``
  It executes args with `os/spawn` and returns a struct which has the following keys.

  * `:out` - the standard output of the process
  * `:err` - the standard error of the process
  * `:exit-code` - the exit code of the process`

  If there is any error, to prevent zombie processes, the spawned process is killed.

  Regardless of errors, the spawned process is going to be closed for resource control.
  ``
  [& args]
  (with [proc (os/spawn args :p {:out :pipe :err :pipe})]
    (try
      (let [[out err ec]
            (ev/gather
              (ev/read (proc :out) :all)
              (ev/read (proc :err) :all)
              (os/proc-wait proc))]
        {:out out :err err :exit-code ec})
      ([err f]
        (try
          (os/proc-kill proc)
          ([_]))
        (propagate err f)))))
```

I'm a bit worried about pipe buffer. If `ev/with-deadline` detaches the functions above from a zombie process, `ev/read` may not fully drain the pipe buffer. Does `os/proc-kill` drain pipe buffer?

bakpakin:
If you are worried about draining the pipe buffer in case of timeout, I recommend first setting up a buffer for stdout and start ahead of time, and then replacing (ev/read out :all) with (ev/read out :all my-buffer). That works for me to see the output of Ls even after timeout

[...]

PR #174, opened 2024-02-03
Tighter resource control for `sh/exec-slurp(-all)`. Add `stream/lines(-channel)`.

amano-kenji:
the processes spawned by `sh/exec-slurp(-all)` are properly closed for resource control.

This closes https://github.com/janet-lang/spork/issues/173

`stream/lines` returns a fiber that yields lines from a stream. `stream/lines-channel` returns a channel that gives lines from a stream.


amano-kenji:
I'm working on lower-level `sh/line` functions that expose `core/process`.

amano-kenji:
I think it's ready.

sogaiu:
@tionis May be you would be interested in taking a look at the proposed changes in this PR?

amano-kenji:
It seems perfectly functional now, but I wonder whether the first message from a channel returned by `lines` should be the function that cancels the fiber that feeds the channel and kills the command process.

A struct as the return value feels a bit clunky.

tionis:
> @tionis May be you would be interested in taking a look at the proposed changes in this PR?

Looks fine to me.
Why has the docstring so many blank lines, though?

amano-kenji:
They are paragraphs....

amano-kenji:
I think I should give frozen processes a chance to shut down gracefully before terminating them forcefully. I'm going to forcefully push soon.

tionis:
> They are paragraphs, but I think I can combine them into fewer paragraphs.

That might look cleaner, be more readable in the repl.

> I think I should give frozen processes a chance to shut down gracefully before terminating them forcefully. I'm going to forcefully push soon.

This seems reasonable, the timeout should probably be configurable either via args (should be possible by walking the argument list as until now only strings where valid arguments) or dynamic variable from the environment (more idiomatic and performant but uses a global var)

amano-kenji:
I think it is ready for review. It is cleaner now. I don't actually have to allow users to configure timeout because the frozen processes are killed asynchronously in a janet task. The timeout is 30 seconds which should be enough.

amano-kenji:
My little worry is that another process takes over the PID that was killed with SIGTERM, and `kill-proc-async` ends up killing it with SIGKILL.

sogaiu:
A bit of a hack perhaps, but on Linux, perhaps determining the start time of the process adapting [this idea](https://serverfault.com/questions/95643/linux-how-do-i-see-when-a-process-started/95646#95646) [1] and then acting or not acting accordingly might help?

---

[1] So parsing `/proc/$pid/stat` for appropriate info...

amano-kenji:
Is `/proc/$pid/stat` specific to linux?

amano-kenji:
FreeBSD also has `/proc/$pid/stat`. In both FreeBSD and Linux, starttime has the same position in `/proc/$pid/stat`, but `/proc/$pid/stat` is not the same between BSD and linux.

I wish janet proc had starttime.

amano-kenji:
I think the only portable way to handle this is to kill proc only once with SIGTERM or SIGKILL. SIGKILL doesn't leave a trail, but it doesn't allow proc to shut down gracefully. SIGTERM allows it to shut down gracefully, but it could leave a trail in some cases.

amano-kenji:
Because I cannot call `os/proc-wait` twice or even three times, my options are limited. I guess I'm going to go ahead with `SIGTERM` and remove `:x` flag in `os/spawn` for better handling of non-zero exit code.

amano-kenji:
After many iterations, I realized killing processes in `sh/exec-slurp-(all)` isn't a great idea.

After the pipewire freezing issue was fixed, I realized the freezing issue has to be addressed in upstream projects, not in `spork/sh`. Addressing the symptom of an underlying issue is a bad idea. The issues have to be fixed at the root causes.

Thus, I made the pull request simpler. Now, `sh/exec-slurp` and `sh/exec-slurp-all` don't try to kill processes upon errors. They just try to close process pipes.

`sh/lines` also just simply try to send SIGTERM to the underlying process when `:cancel` function is called.

tionis:
This does seem cleaner and with the removed risk of unexpected killing of processes :+1:

amano-kenji:
I decided not to make `sh/slurp-exec` and `sh/slurp-exec-all` responsible for killing the spawned program.

I just made `:cancel` function for `sh/lines` handle errors more gracefully. I don't want to make further modifications unless they are necessary.

Ready for review and merge.

amano-kenji:
I believe It's a correct behavior for `sh/slurp-exec` and `sh/slurp-exec-all` to not kill their processes upon errors.

If I cancel `os/execute` with `ev/with-deadline`, `os/execute` doesn't kill its process.

I can think of `sh/slurp-exec` and `sh/slurp-exec-all` as `os/execute`. Users who want more control should use `os/spawn` directly.

amano-kenji:
I think I finally streamlined this pull request...

sogaiu:
Very minor point, but is [this do](https://github.com/janet-lang/spork/pull/174/files#diff-e45d6fff2a878fae71bd94a45a1f2277a2b015d8b271ef6437d147e136009a80R53) needed?

bakpakin:
LGTM, thanks for all the work on this, @amano-kenji.

@sogaiu I'm not to concerned with minor style points on little things, esp. since this should generate identical bytecode.

amano-kenji:
> Very minor point, but is [this do](https://github.com/janet-lang/spork/pull/174/files#diff-e45d6fff2a878fae71bd94a45a1f2277a2b015d8b271ef6437d147e136009a80R53) needed?

That is a minor mistake...... I didn't spot that....

sogaiu:
Seem to be no harm done :+1:

amano-kenji:
Thanks, bakpakin.

Issue #189, opened 2024-07-16
Lack of cross-platform function to create temporary directory


sogaiu:
On more than one occasion, I've wanted to create a (previously non-existent) temporary directory to dump some files in.  I'm guessing others have too (^^;

By "cross-platform", I'm mostly meaning various BSDs, Linux, macos, and recent Windows.

Did a bit of digging and have come up with the following links:

* [Temporary folders Wikipedia page](https://en.wikipedia.org/wiki/Temporary_folder)
* [bsds (hier(7)) and linux `/tmp` and `/var/tmp`](https://superuser.com/questions/332610/where-is-the-temporary-directory-in-linux)
* [macos `$TMPDIR`](https://osxdaily.com/2018/08/17/where-temp-folder-mac-access/)
* [windows `$TMP`, `$TEMP`, `$USERPROFILE`,`GetTempPath2`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2a)

Following is a sketch of a possible flow:

1. find "system" / "user" temporary directory
2. verify it exists (it could disappear while doing the following step, but we'll ignore that for now)
3. keep trying to make a new subdirectory of located + verified directory from steps 1 and 2 until success or some number of failures

The focus here is not on hiding things away (like is done in parts of Python's [tempfile](https://docs.python.org/3/library/tempfile.html) library), but just to have some place that won't accidentally get used.

Does this sound like:

* it might work
* something worth having in spork?

sogaiu:
Below is an initial sketch.

```janet
# https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w
(defn windows-temp-root
  []
  (os/getenv "TMP"
             (os/getenv "TEMP"
                        (os/getenv "USERPROFILE"
                                   # XXX
                                   (os/getenv "WINDIR")))))


# https://en.cppreference.com/w/cpp/filesystem/temp_directory_path
(defn posix-temp-root
  []
  (os/getenv "TMPDIR"
             (os/getenv "TMP"
                        (os/getenv "TEMP"
                                   (os/getenv "TEMPDIR" "/tmp")))))

(defn temp-root
  []
  (case (os/which)
    # see comment above function definition
    :windows (windows-temp-root)
    # XXX: unsure
    :mingw "/tmp"
    # XXX: unsure, but https://cygwin.com/cygwin-ug-net/setup-env.html
    :cygwin "/tmp"
[...]

PR #190, opened 2024-07-19
Add make-temp-dir and friends

sogaiu:
This is an attempt to address #189 including some tests.

There are four functions:

* `make-temp-dir`
* `temp-root`
* `windows-temp-root`
* `posix-temp-root`

`make-temp-dir` tries to create a temporary directory that lives under some temporary directory that a system already has (the function `temp-root` tries to determine this) and tries to do so according to a template.

`windows-temp-root` and `posix-temp-root` are supporting functions for the `temp-root` function.

See the docstring of each for details.

I'm not certain what `temp-root` should return for MinGW and Cygwin setups.  For the moment I've gone with `/tmp` for both.  If someone happens to know or can check, I'm happy to modify.  Otherwise, may be if these are wrong, they can get corrected later (^^;

### DIFF
--- a/spork/sh.janet
+++ b/spork/sh.janet
@@ -171,3 +171,95 @@
   "Output a string with all arguments correctly quoted"
   [& args]
   (string/join (map shell-quote args) " "))
+
+# https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w
+(defn windows-temp-root
+  ``
+  Try to determine a temporary directory for a Windows system.
+  ``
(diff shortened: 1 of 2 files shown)


sogaiu:
Not making much progress getting info about the MinGW and Cygwin cases and may be it wouldn't be bad to get some more experience using the existing code so I'm going to close this for now.

---

On a related note, @yumaikas has a library called [janet-tempfiles](https://github.com/yumaikas/janet-tempfiles) that has some intersecting functionality.

PR #193, opened 2024-08-22
Reorganize tests to be one file per module and to give more descriptive file names

CFiggers:
Historically, spork's tests have been organized into numbered suites, such as `suite0002.janet`. Each suite usually contains tests related to a single module. But some suites mix together tests for unrelated modules (without an apparent intention to doing so).

More recent additions and updates in the `spork/test` directory have switched to using an (IMHO) much friendlier convention. These newer test files simply name the test file after the module it is primarily focused on testing, prefixed with `suite-`. This makes test cases for a given module much easier to find. Discoverability of test cases matters because **in many cases the tests are the only documentation or usage examples available for a given spork module.**

This PR proposes renaming and lightly reorganizing spork's tests to entirely switch over to the more recently-preferred convention. The goal is to improve the discoverability and therefore usefulness of test suites as quasi-documentation. The tests themselves are unmodified and all still pass.


sogaiu:
I tried your branch and tallied the number of tests before and after and the counts seem to match.

LGTM :)

PR #199, opened 2024-09-22
Add capture* functions to temple module

### LINKED ISSUE #186: Capturing functions for the temple
In the time of writing, when one imports temple `.template`, the returned environment consists of two functions: `render` and `render-dict`, which is in line with the language accent on printing to stdout. But when you do not use this approach in rendering (as the `httpf` module of this very library), you have to litter your code with `capout` calls. I used macro, which will import the library, render it, and capture the stdout, but it does not play well with flycheck lint and janet-lsp server.

I propose to have two more functions returned in the environment, analogous to the two existing ones, but returning the captured string:
- `capture` analog to `render`
- `capture-dict` analog to `render-dict`

I am aware of the `compile` function, but I really like the process of importing templates and then calling the functions. 

Also, I will volunteer for the PR if the change is deemed useful.

pepe:
Closes #186

### DIFF
--- a/spork/temple.janet
+++ b/spork/temple.janet
@@ -127,21 +127,25 @@
   [path &]
   (with-dyns [:current-file path]
     (let [tmpl (create (slurp path) path)]
-      @{'render @{:doc "Main template function."
-                  :value (fn render [&keys args] (tmpl args))}
-        'render-dict @{:doc "Template function, but pass arguments as a dictionary."
-                       :value tmpl}
-        'capture @{:doc "Template function that returns buffer of rendered template."
-                   :value (fn render [&keys args]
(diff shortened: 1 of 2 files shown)


pepe:
The functions are not renamed, I will fix it tomorrow 😉

sogaiu:
Good to see there is some evidence of testing :)

pepe:
I did it in the last ten conscious minutes of my day yesterday:-D. Then, already in bed, it came to my mind that just copy-paste is not that great, so I checked it and made the comment. 

When I get back home, I will fix the names.

pepe:
I guess, it can be checked now.

PR #202, opened 2024-11-12
Fix reading text/stream content type

pepe:
Also remove unused head-size and format the module code.


pepe:
I use `http/request` in my tests, and I found this omission when using DataStar's SSE.

PR #203, opened 2024-11-25
Advertise client capability of janet-netrepl

sogaiu:
Currently `janet-netrepl`'s client capability is slightly under-promoted.

This PR is an attempt to mititgate this a bit.

PR #204, opened 2024-11-25
Define symbol early enough for use in math

sogaiu:
There was [a CI error related to #203](https://github.com/janet-lang/spork/actions/runs/12004487043/job/33459550150?pr=203).

My current guess is that due to [recent changes in hashing in janet](https://github.com/janet-lang/janet/issues/1520), the order in which the key-value pairs are evaluated has changed.

So before may be: https://github.com/janet-lang/spork/blob/44e26016352a66fd49470eb2170aa111d887484b/spork/math.janet#L360-L363

worked fine because luckily the `(def m ...)` happened before the use of `m` for the value associated with `:b`.

This PR just moves the `def` form out of the struct.

### DIFF
--- a/spork/math.janet
+++ b/spork/math.janet
@@ -357,9 +357,10 @@
         (+= sy y)
         (+= sxx (* x x))
         (+= sxy (* x y)))
-      {:m (def m
-            (/ (- (* cl sxy) (* sx sy))
-               (- (* cl sxx) (* sx sx))))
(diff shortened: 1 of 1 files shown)


sogaiu:
I'm not entirely clear on why the CI is failing.

FWIW, locally, the PR content appears to work fine, i.e. `jpm test` leads to no errors.

PR #220, opened 2025-06-09
Handle windows UNC and namespaced paths in path/normalize function

agent-kilo:
Currently `path/win32/normalize` removes all duplicate backslashes (`\`) in front of a path, but double backslashes (`\\`) in front of a path name means to access other "namespaced" paths in Windows. This PR tries to modify `path/win32/normalize` function so that it handles paths starting with `\\`.

Specifically, the code tries to handle these cases:

* `\\wsl$\dist\path`: A path pointing to a file inside WSL
* `\\server\share\path`: A path pointing to a file in a shared folder on a remote server (UNC path)
* `\\?\path\to\file` and `\\.\path\to\file`: DOS device paths

More info on [learn.microsoft.com](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats) and [wikipedia](https://en.wikipedia.org/wiki/Path_(computing)#Universal_Naming_Convention).


pepe (approved):
LGTM, thank you!

PR #226, opened 2025-07-08
Fix use of `distinct` on macOS lflags with `-framework`

rwtolbert:
Summary: Current code uses `distinct` to generate a unique list off lflags for linking executables. This fails on macOS when linking to a framework with flags like:

["-Lfoo" "-lbar" "-lfoo" "-framework" "Cocoa" "-framework" "OpenGL"]

`distinct` generates:

@["-Lfoo" "-lbar" "-lfoo" "-framework" "Cocoa" "OpenGL"]

where `-framework` is missing for OpenGL.

### DIFF
--- a/spork/declare-cc.janet
+++ b/spork/declare-cc.janet
@@ -804,11 +804,31 @@ int main(int argc, const char **argv) {
               false)
             static))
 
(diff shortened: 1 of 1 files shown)


bakpakin:
LGTM!

PR #178, opened 2024-02-16
CI/CD feature: Remove "This workflow requires approval from a maintainer"

hyiltiz:
Removes friction from PR contributions and automatically triggers the linters/checks without first getting "approval" from repo maintainer.

PR #229, opened 2025-08-14
Brief: Add empty rules for pre-clean/post-clean

rwtolbert:
Summary: The previous fix to add all these new rules did not check for the case where no pre-clean/post-clean was defined in a project.

as such, running `janet-pm clean` without defining `pre-clean` in the project.janet resulted in an error. The original tests all passed since all pre/post steps were defined.

This fix adds default empty rules
for each and then tests this in the test-bundle project that doesn't define pre-clean/post-clean.

### DIFF
--- a/test/suite-pm-pre-post.janet
+++ b/test/suite-pm-pre-post.janet
@@ -43,6 +43,8 @@
       (each line err
         (printf line)))))
 
+(def bat (if (= (os/which) :windows) ".bat" ""))
+
 (defn- test-binscript [filename syscount printcount]
   (assert (sh/exists? filename))
   (let [contents (slurp filename)]
@@ -53,7 +55,7 @@
     (assert (= syscount (length (string/find-all "put root-env :install-time-syspath" contents))))
     (assert (= printcount (length (string/find-all "print" contents)))))
   # and finally make sure the script actually runs
-  (assert (= "hello" (sh/exec-slurp filename))))
+  (assert (= "hello" (sh/exec-slurp (string filename bat)))))
 
 # Create a temporary directory for our janet tree
 (math/seedrandom (os/cryptorand 16))
@@ -80,16 +82,15 @@
        # verify all bin scripts have two copies of hard-coded syspath, one in body
        # and one in main
        (def binpath (path/join (dyn *syspath*) "bin"))
-       (def exe (if (= (os/which) :windows) ".exe" ""))
-       (def janet-pm (path/join binpath (string "janet-pm" exe)))
+       (def janet-pm (path/join binpath (string "janet-pm")))
        (assert (sh/exists? janet-pm))
        (assert (= 2 (length (string/find-all "put root-env :install-time-syspath" (slurp janet-pm)))))
 
-       (def janet-format (path/join binpath (string "janet-format" exe)))
+       (def janet-format (path/join binpath (string "janet-format")))
        (assert (sh/exists? janet-format))
        (assert (= 2 (length (string/find-all "put root-env :install-time-syspath" (slurp janet-format)))))
 
-       (def janet-netrepl (path/join binpath (string "janet-netrepl" exe)))
+       (def janet-netrepl (path/join binpath (string "janet-netrepl")))
        (assert (sh/exists? janet-netrepl))
        (assert (= 2 (length (string/find-all "put root-env :install-time-syspath" (slurp janet-netrepl)))))
 
@@ -97,7 +98,7 @@
 
        (os/cd "test-bundle")
        (divider "Running janet-pm clean in a project that doesn't define pre-clean/post-clean")
-       (def empty-out (sh/exec-slurp-all janet-pm "clean"))
+       (def empty-out (sh/exec-slurp-all (string janet-pm bat) "clean"))
        (dump-out empty-out)
        (assert (= 0 (count-match "error" (empty-out :err))))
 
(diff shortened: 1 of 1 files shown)


rwtolbert:
I'm closing in on another patch. When calling `Janet-pm` in my tests, it is ignoring the hard-coded syspath in the top of the file and then trying to install into the system, which of course it can't without permission.

I'm trying a fix that puts the hard-coded sys path inside the `main` where they do appear to affect the install location.

pyrmont:
@rwtolbert You’re not going crazy. The issue seems to be due to the GitHub workflow using `pull_request_target`:

https://github.com/janet-lang/spork/blob/6f189ab0458075ac23f0d3306c8e9c870829ffe2/.github/workflows/test.yml#L3-L7

Based on [this SO answer](https://stackoverflow.com/a/74959635/308909), this causes the test to run by checking out the commit of the target (i.e. the current HEAD of the master branch) rather than a merge commit that would become the current HEAD of the master branch.

The change was added in 05eb0140db629fa1eb6de0fdbfbb45cedec32d04, apparently as a way to get the workflow to run without needing a maintainer to approve it (see #178). If I’m understanding things correctly, **this means that pull request checks are currently meaningless** because all it’s doing is running the current HEAD of master. 

As for the issue that #178 identified, I don’t have commit privileges for Spork so I’m not sure what its settings are but in the repos I run, I have the security setting in Settings > Actions set to ‘Require approval for first time contributors’. If that’s similarly set with Spork and it’s considered overly limiting, @bakpakin could set it to ‘Require approval for first-time contributors who are new to GitHub’.

My view is that requiring approval for first-time contributors is a reasonable default and not an excessive amount of friction. In other words, I would revert the change in 05eb0140db629fa1eb6de0fdbfbb45cedec32d04.

pyrmont:
I don't think you need to feel bad. I assume the reason the checks that are causing a problem now because they are running against the current HEAD is the same reason that the checks in that PR didn't catch the error. Just a mistake in the workflow file!

rwtolbert:
Very true. and I appreciate the sentiment. But it drives me crazy to have broken code - especially when I've done the breaking.

And from what I can tell, there is no way I can get this to work from my end.

rwtolbert:
and perhaps this is currently moot, since the tests don't run on Windows. I just got access to my Windows machine and I'm working through the `sh/exec` stuff in the tests.

sogaiu:
I tried https://github.com/janet-lang/spork/pull/229/commits/ef0f84206078d08e0c8ee075eefa220fe0808a4b.

It looks like `jpm test` won't work without `spork` already being installed?

```
running test/suite-pm-pre-post.janet ...
executable scripts have been installed to /home/user/src/spork.rwtolbert/tmp/tmp_dir_0894973558081855.tmp/bin
  in require-1 [boot.janet] on line 3126, column 20
  in import* [boot.janet] on line 3170, column 15
  in thunk [/home/user/src/spork.rwtolbert/tmp/tmp_dir_0894973558081855.tmp/bin/bin-no-hardcode] (tail call) on line 3, column 1
error: command failed with non-zero exit code 1
  in os/proc-wait [src/core/os.c] on line 647
  in ev/gather [spork/sh.janet] on line 34, column 19
  in defer [boot.janet] on line 3900, column 13
  in wait-for-fibers [boot.janet] on line 3893, column 5
  in defer [spork/sh.janet] on line 32, column 17
  in exec-slurp [spork/sh.janet] on line 31, column 3
  in test-binscript [test/suite-pm-pre-post.janet] on line 58, column 22
  in defer [test/suite-pm-pre-post.janet] on line 149, column 8
  in thunk [test/suite-pm-pre-post.janet] (tail call) on line 68, column 1
non-zero exit code in test/suite-pm-pre-post.janet: 1
```

bakpakin:
Thanks for all the effort here, @rwtolbert . In the meantime, I am reverting the last merge to get builds working again. I haven't taken a closer look but may find some time this weekend to see if I can help. There does seem to be some CI weirdness going on, which should separately be investigated and improved.

sogaiu:
Did some testing on Linux for b25738112f5fcdff504a2d89bc6ea4a6d2f96092 (has had two "reverts") and what I think is the equivalent bcad982480a0d26fc1bba3e2b63e86c27ff1d502?).

---

Good news: `jpm test` works without installation in both.

---

Other news: `SPORK_TEST_ALL_PACKAGES=1 jpm test` seems to be having some kind of issue in both (output truncated below):

```
remote: Enumerating objects: 161, done.
remote: Counting objects: 100% (30/30), done.
remote: Compressing objects: 100% (24/24), done.
remote: Total 161 (delta 10), reused 8 (delta 5), pack-reused 131 (from 1)
Receiving objects: 100% (161/161), 3.77 MiB | 10.19 MiB/s, done.
Resolving deltas: 100% (88/88), done.
From https://github.com/janet-lang/sqlite3
 * [new branch]      master     -> origin/master
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0 (from 0)
From https://github.com/janet-lang/sqlite3
 * branch            HEAD       -> FETCH_HEAD
HEAD is now at a19ec48 Merge pull request #20 from talmeme/using_extensions
error: project.janet:37:1: compile error: unknown symbol body
  in dofile [boot.janet] on line 3102, column 7
  in thunk [/tmp/tmp_dir_899849061149376.tmp/bundle/sqlite3/init.janet] (tail call) on line 4, column 1
  in dofile [boot.janet] (tail call) on line 3101, column 7
  in source-loader [boot.janet] on line 3113, column 15
  in require-1 [boot.janet] (tail call) on line 3137, column 18
  in defer [boot.janet] on line 4200, column 7
  in get-bundle-module [boot.janet] on line 4192, column 5
  in edefer [boot.janet] on line 4351, column 19
  in bundle/install [boot.janet] (tail call) on line 4339, column 5
  in pm-install [spork/pm.janet] on line 313, column 5
  in pm-install [spork/pm.janet] on line 300, column 7
  in defer [test/suite-pmfull.janet] on line 42, column 5
  in thunk [test/suite-pmfull.janet] (tail call) on line 33, column 3
non-zero exit code in test/suite-pmfull.janet: 1
```

PR #230, opened 2025-08-16
Try 2: add more pre/post steps for janet-pm

### LINKED ISSUE #225: janet-pm runs `(prebuild)` before all custom tasks
I like the ability to add pre-build tasks so that I can build some C++ dependencies before the actual `build` step with just `janet-pm build`

However, it seems `janet-pm` runs `(prebuild)` before every task, even clean and clean-all.

In the worst case, if I run `janet-pm clean-all` twice, the first invocation will remove the built C++ dependencies as part of cleaning which is what I want. But if I run a second `janet-pm clean-all`, it will run the pre-build step, build all the C++ deps, then delete them.

I think the culprit is here:

https://github.com/janet-lang/spork/blob/5601dc883535473bca28351cc6df04ed6c656c65/spork/declare-cc.janet#L308C1-L310C51

```janet
  (defn run-task [task]
    (prebuild)
    (build-rules/build-run e task (dyn :workers)))
```

and if I write new tasks to run via `janet-pm run foo` then prebuild will run before each of them as well.

I can see the usefullness of a new `pre-task` task that is run before everything, but I think `pre-build` should only be run in the `build` hook

rwtolbert:
1. Adding these extra steps allows more user control of the build process for complicated builds, including pre/post steps for build, install, test and clean.

This also fixes #225, so that prebuild is only run before build.

2. Also removes redundant calls to build in the install hook since Janet's core install hook calls build first as well. Having this in spork too resulted in two calls to build when `janet-pm install` was run.

3. Adds local assertf in `janet-pm` for older Janet

4. Fixes existing problem in `janet-pm` where hardcoded syspath at the top-level is not honored when run as `(defn main ...`. This change adds a copy of the hardcoded paths inside defn main in a binscript, if main exists in the input.

### DIFF
--- a/bin/janet-pm
+++ b/bin/janet-pm
@@ -5,6 +5,19 @@
 (import spork/pm)
 (import spork/pm-config)
 (import spork/sh)
+
+# Fix for janet without assertf
+(compwhen (not (dyn 'assertf))
+  (defmacro- assertf
+    "Convenience macro that combines `assert` and `string/format`."
+    [x fmt & args]
+    (def v (gensym))
+    ~(do
+       (def ,v ,x)
+       (if ,v
+         ,v
+         (,errorf ,fmt ,;args)))))
+
 # Do not import declare-cc here - keep a clean separation for faster load times
 
 (def help-text
--- a/spork/declare-cc.janet
+++ b/spork/declare-cc.janet
@@ -266,11 +266,18 @@
   (def bd (build-dir))
   (def rules (get-rules))
   # Initialize build rules
-  (build-rules/build-rule rules :install ["build" "pre-install"])
+  (build-rules/build-rule rules :install [])
   (build-rules/build-rule rules :pre-install [])
+  (build-rules/build-rule rules :post-install [])
   (build-rules/build-rule rules :build [])
   (build-rules/build-rule rules :pre-build [])
+  (build-rules/build-rule rules :post-build [])
+  (build-rules/build-rule rules :check [])
   (build-rules/build-rule rules :pre-check [])
+  (build-rules/build-rule rules :post-check [])
+  (build-rules/build-rule rules :clean [])
+  (build-rules/build-rule rules :pre-clean [])
+  (build-rules/build-rule rules :post-clean [])
   # Add hooks
   (def e (curenv))
   (defn- prebuild
@@ -279,34 +286,59 @@
     (os/mkdir bd)
     (os/mkdir (path/join bd "static"))
     (build-rules/build-run e "pre-build" (dyn :workers)))
+  (defn- postbuild
+    []
+    (build-rules/build-run e "post-build" (dyn :workers)))
   (defn- precheck
     []
     (build-rules/build-run e "pre-check" (dyn :workers)))
+  (defn- postcheck
+    []
+    (build-rules/build-run e "post-check" (dyn :workers)))
+  (defn- preinstall
+    []
+    (build-rules/build-run e "pre-install" (dyn :workers)))
+  (defn- postinstall
+    []
+    (build-rules/build-run e "post-install" (dyn :workers)))
+  (defn- preclean
+    []
+    (build-rules/build-run e "pre-clean" (dyn :workers)))
+  (defn- postclean
+    []
+    (build-rules/build-run e "post-clean" (dyn :workers)))
   (defn build [&opt man target]
     (prebuild)
     (default target "build")
-    (build-rules/build-run e target (dyn :workers)))
+    (build-rules/build-run e target (dyn :workers))
+    (postbuild))
   (defn install [manifest &]
-    (build)
+    # (build) - removed since install in janet/src/boot/boot.janet calls build in the install hook
+    (preinstall)
     (with-dyns [*install-manifest* manifest]
-      (build-rules/build-run e "install" (dyn :workers))))
+      (build-rules/build-run e "install" (dyn :workers)))
+    (postinstall))
   (defn check [&]
     (build)
     (precheck)
-    (run-tests))
+    (run-tests)
+    (postcheck))
   (defn list-rules [&]
     (each k (sorted (filter string? (keys rules)))
       (print k)))
   (defn rule-tree [&]
     (show-rule-tree rules))
   (defn clean [&]
+    (preclean)
     (print "removing directory " bd)
-    (sh/rm bd))
+    (sh/rm bd)
+    (postclean))
   (defn clean-all [&]
+    (preclean)
     (print "removing directory " br)
-    (sh/rm br))
+    (sh/rm br)
+    (postclean))
   (defn run-task [task]
-    (prebuild)
     (build-rules/build-run e task (dyn :workers)))
   (defglobal 'install install)
   (defglobal 'build build)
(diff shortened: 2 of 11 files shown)


sogaiu:
I gave https://github.com/janet-lang/spork/pull/230/commits/c9f9a5806e3e8add52a246fc2f86b0d0daea2097 a try on a Linux box.

`jpm test` works once I have `spork` installed :+1: 

---

Without `spork` installed, I see some error output in the `jpm test` output.

Below is some of the output:

```
running test/suite-pm-pre-post.janet ...
executable scripts have been installed to /home/user/src/spork.rwtolbert/tmp/tmp_dir_469074351512296.tmp/bin
  in require-1 [boot.janet] on line 3126, column 20
  in import* [boot.janet] on line 3170, column 15
  in thunk [/home/user/src/spork.rwtolbert/tmp/tmp_dir_469074351512296.tmp/bin/bin-no-hardcode] (tail call) on line 3, column 1
error: command failed with non-zero exit code 1
  in os/proc-wait [src/core/os.c] on line 647
  in ev/gather [spork/sh.janet] on line 34, column 19
  in defer [boot.janet] on line 3900, column 13
  in wait-for-fibers [boot.janet] on line 3893, column 5
  in defer [spork/sh.janet] on line 32, column 17
  in exec-slurp [spork/sh.janet] on line 31, column 3
  in test-binscript [test/suite-pm-pre-post.janet] on line 58, column 22
  in defer [test/suite-pm-pre-post.janet] on line 149, column 8
  in thunk [test/suite-pm-pre-post.janet] (tail call) on line 68, column 1
non-zero exit code in test/suite-pm-pre-post.janet: 1
```

---

Also, `SPORK_TEST_ALL_PACKAGES=1 jpm test` still gives error output here even if spork is already installed -- may be that's a separate issue (or pilot error)?

In any case, below is some of the associated output:

```
Initialized empty Git repository in /tmp/tmp_dir_678199816948947.tmp/.cache/git__https___github.com_pyrmont_testament/.git/
remote: Enumerating objects: 456, done.
remote: Counting objects: 100% (176/176), done.
remote: Compressing objects: 100% (38/38), done.
remote: Total 456 (delta 145), reused 150 (delta 130), pack-reused 280 (from 1)
Receiving objects: 100% (456/456), 108.90 KiB | 9.90 MiB/s, done.
Resolving deltas: 100% (221/221), done.
From https://github.com/pyrmont/testament
 * [new branch]      master     -> origin/master
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0 (from 0)
From https://github.com/pyrmont/testament
 * branch            HEAD       -> FETCH_HEAD
HEAD is now at a434abe Support dynamic bindings :tests and :skips to call/skip tests
error: assert failure in (dyn *install-manifest*)
  in pre-install [/tmp/tmp_dir_678199816948947.tmp/spork/declare-cc.janet] on line 363, column 25
  in edefer [/tmp/tmp_dir_678199816948947.tmp/spork/build-rules.janet] on line 118, column 22
  in _while [/tmp/tmp_dir_678199816948947.tmp/spork/build-rules.janet] on line 112, column 7
  in worker [/tmp/tmp_dir_678199816948947.tmp/spork/build-rules.janet] on line 105, column 5
  in defer [/tmp/tmp_dir_678199816948947.tmp/spork/build-rules.janet] on line 19, column 11
  in wait-for-fibers [/tmp/tmp_dir_678199816948947.tmp/spork/build-rules.janet] on line 12, column 3
  in run-rules [/tmp/tmp_dir_678199816948947.tmp/spork/build-rules.janet] (tail call) on line 132, column 3
  in install [/tmp/tmp_dir_678199816948947.tmp/spork/declare-cc.janet] (tail call) on line 317, column 5
  in do-hook [boot.janet] on line 4210, column 5
  in edefer [boot.janet] on line 4358, column 7
  in bundle/install [boot.janet] (tail call) on line 4339, column 5
  in pm-install [spork/pm.janet] on line 313, column 5
  in pm-install [spork/pm.janet] on line 300, column 7
  in pm-install [spork/pm.janet] on line 300, column 7
  in defer [test/suite-pmfull.janet] on line 42, column 5
  in thunk [test/suite-pmfull.janet] (tail call) on line 33, column 3
non-zero exit code in test/suite-pmfull.janet: 1
```

rwtolbert:
@sogaiu Thanks for the pointer to `SPORK_TEST_ALL_PACKAGES=1`. I've not seen that  before and I guess the CI tests don't use it.

Will have to dig in a bit more.

sogaiu:
Regarding this peg:

```janet
{:ws (some (choice :a :d :s "[" "&"))
 :defn (sequence "(defn" (any "-") :s "main" :ws "]" (any (choice " " "\n")))
 :main :defn}
```

Is it correct that its purpose is to try to match some initial portion of the definition of a `main` function, e.g.

```janet
(defn main
  [& args]
```

?

sogaiu:
> Will have to dig in a bit more.

I think it may be a problem that existed before this PR -- I think I saw it happen in https://github.com/janet-lang/spork/commit/b25738112f5fcdff504a2d89bc6ea4a6d2f96092 and https://github.com/janet-lang/spork/commit/bcad982480a0d26fc1bba3e2b63e86c27ff1d502 when [checking recently](https://github.com/janet-lang/spork/pull/229#issuecomment-3193323334).

rwtolbert:
@sogaiu That PEG is designed (hopefully) to find a `main` in a binscript so that janet-pm can insert the hardcoded syspath inside/at the top of the  main.

The current behavior inserts them at the top of the file, below the shebang line, but these variables are NOT available/honored when the script is executed as a `main` 

it isn't the prettiest peg and I see now that `:defn` could just be `:main`

sogaiu:
> to find a main in a binscript so that janet-pm can insert the hardcoded syspath inside/at the top of the main.

I guess it is kind of tricky to do this well because in general the content of the binscript is likely written by other folks (e.g. there could be destructuring in the argument tuple, metadata, docstring, etc.).

FWIW, I maintain [some pegs that parse janet source](https://github.com/sogaiu/janet-peg) that I've been using as part of manipulating janet source (and there's something similar [in janet's tests](https://github.com/janet-lang/janet/blob/bf343407373bcec01e236f328fc7d8022da69110/test/suite-peg.janet#L366-L403) as well as [in the docs](https://janet-lang.org/docs/syntax.html#grammar))...if this approach is what is going to be followed, perhaps it's worth trying to add something appropriate to spork and use that?

I wonder if @bakpakin (or anyone else) has any ideas about potential alternative approaches though...

rwtolbert:
@sogaiu I agree with your comments about "finding main". I think the PEG and code are safe such that if it doesn't find anything, it just passes the existing code as-is. 

and I hate to have to do it, but otherwise I can't see any other way for those hard-coded paths to work in a script that is run via main. It could also be that most end-user binscripts wouldn't care, but certainly `janet-pm` needs it. We could even write it very specifically only for `janet-pm`

rwtolbert:
this may just be the Janet corollary to the old regex joke.

"You have a problem and you think, I know, I'll use a PEG. Now you have two problems"

rwtolbert:
@sogaiu Yes, I think you are right to remove it.

PR #240, opened 2025-09-16
Fix xcopy invocation in sh/copy

pepe:
Separated parts with a space and added /-i switch, which turns off the prompt by the xcopy command. I encountered this bug while trying to create the quickbin with `spork/declare-cc`, and it stuck.


pepe:
I need to test it on the POSIX. Converted to draft for the moment.

sogaiu:
FWIW, with https://github.com/janet-lang/spork/pull/240/commits/bc1744c26bd0a7334d21dd42df4b36afcda06a88, both `jpm test` and `janet-pm test` gave no errors :+1:

PR #241, opened 2025-09-17
Fix PATH separator in ps activate script

pepe:
Another POSIX muscle memory catch. I am enjoying the usage of special environments a lot!


sogaiu:
LGTM

PR #242, opened 2025-09-17
add some jpm sub-commands to janet-pm. 

rwtolbert:
Code is already there in `declare-cc.janet` and in `bundle`, this just adds the options to janet-pm.

### DIFF
--- a/bin/janet-pm
+++ b/bin/janet-pm
@@ -111,10 +111,14 @@
     load-lockfile (src)
         Install all bundles in the given lockfile.
 
+    run rule
+        Run a rule. Can also run custom rules added via `(task "task"
+        [deps...] ...)` or `(rule "output.file" [deps...] ...)`.
+
     rules
         List rules available with run.
 
-    rule-tree (root rule) (depth)
+    rule-tree (root) (depth)
         Print a nice tree to see what rules depend on other rules.
         Optionally provide a root rule to start printing from, and a
         max depth to print. Without these options, all rules will
--- a/bin/janet-pm
+++ b/bin/janet-pm
@@ -21,7 +21,7 @@
 # Do not import declare-cc here - keep a clean separation for faster load times
 
 (def help-text
-    ```
+  ```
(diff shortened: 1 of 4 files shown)


sogaiu:
The three subcommands added seem to work fine here :+1: 

About some of the docs...

>     rules
>        List rules available with run.

There is a reference to `run` -- which I think `jpm` does have.  Does `janet-pm` have a `run` subcommand though?

I'm not seeing one in the current version nor in this PR.

May be I'm missing something :sweat_smile:

sogaiu:
Thanks!

I gave https://github.com/janet-lang/spork/pull/242/commits/9f828fa135d55c54aa757396fe93fed268090820 a try.

One thing I noticed is that `janet-pm build` and `janet-pm run build` do not seem to behave the same.

I think `jpm build` and `jpm run build` do (though I may be wrong).

Do you know if the `janet-pm` behavior is intentional?

rwtolbert:
This difference in behavior is definitely a bug/design flaw, but I'm still trying to figure out a fix (if there is one).

Basically, `janet-pm build` runs a hook named `build`, defined by `declare-project`:

```janet
  (defn build [&opt man target]
    (prebuild)
    (default target "build")
    (build-rules/build-run e target (dyn :workers))
    (postbuild))
```

where as, as best I can tell, `janet-pm run build`, runs a specific `build` rule, which skips the `prebuild` step and immediately tries to build all the dependencies of the `build` rule. So `(prebuild)` is not run and the `_build` directory is not created.

if instead, I have `janet-pm run` call `do-hook`, I can get `janet-pm run build` to behave the same as `janet-pm build`

```janet
(defn do-hook
  [hook & args]
  (pm/local-hook hook @{} ;args))
```

notice the extra args passed to the `build` hook function.

but this fix doesn't help `install` and `uninstall` so more investigation is needed.

This probably explains with the `janet-pm run` subcommand was not exposed.

rwtolbert:
This requires a much deeper change. and I suspect that will be tricky to avoid breaking some other behavior.

Closing for now and maybe I can re-open a new one if I figure it out.

PR #251, opened 2025-10-06
Add test for the sh/copy

pepe:


pepe:
I apologize for not providing sufficient context. I have been doing it in a hurry. This is just a test from #240 to see how the current sh/copy implementation copes with it, as can be seen, not even windows-latest does. From my limited understanding, there is no noninteractive way to `cp -r` on Windows 10.

bakpakin:
Closing, this has been merged manually combined with some other fixes.

PR #253, opened 2025-11-05
New module: `spork/date`

CFiggers:
This PR provides a new module, `spork/date`, which contains utility wrappers around `os/time` and `os/date` for working with dates, particularly converting them to and from strings.

This module is motivated by working with SQLite databases (using [janet-lang/sqlite3](https://github.com/janet-lang/sqlite3)). SQLite databases store dates as strings, necessitating both reading dates from strings and writing dates to strings.

This module enables code like the following:

```janet
(import sqlite3)
(import spork/date)

(def db (sqlite3/open "db.db"))

(def rows (sqlite3/eval db "select * from my_table")) 
# @[{:date "2025-11-06 12:08:57.000" :id 1 :value 123} 
#   {:date "2025-10-31 12:09:16.000" :id 2 :value 345} 
#   {:date "2025-11-21 12:09:34.000" :id 3 :value 567}]

(each row rows
  (sqlite3/eval db
    "update my_table set date = :date where id = :id"
    {:date (-> (row :date)
               (date/from-string "yyyy-MM-dd hh:mm:ss.000")
               (date/add :days 1)
               (date/to-string "yyyy-MM-dd hh:mm:ss.000"))
     :id (row :id)}))

(def updated-rows (sqlite3/eval db "select * from my_table"))  
# @[{:date "2025-11-07 12:08:57.000" :id 1 :value 123} 
#   {:date "2025-11-01 12:09:16.000" :id 2 :value 345} 
#   {:date "2025-11-22 12:09:34.000" :id 3 :value 567}]
```

I think this is pretty convenient.

Lots of tests included. All exported functions have docstrings. Doesn't break anything existing.

Feedback welcome.


sogaiu:
As another user of the sqlite3 bits, thank you :)

Some minor initial comments follow.

---

* I'm not a fan of using `use` usually, but if one were to use `use`, would there not be a collision with janet's built-in `compare`?

* How about `short-months` and `long-months` instead of `short-month-map` and `long-month-map` as names?  The `-map` felt slightly misleading.  I understand these are internal names, but may be it's less confusing for first-time (and returning?) readers?

* Is it worth exposing `leap-year?` may be?

* The docstrings for `lt` and `gt` don't seem to make it clear that more than two dates can be passed in, is that intentional?

* Could it be worth constructing `my-format-peg` outside of `to-string`?  It doesn't seem like it depends on anything passed into `to-string.  With the current code, won't `my-format-peg` be constructed on each call to `to-string` (not sure)?

CFiggers:
Yeah, it's a bug in the `spork/test/assert-docs` function—if you provide a parenthesised form on the first line of a docstring (as I just added for `date?` and `assert-date`) it expects that line to be followed by a double newline (`\n\n` specifically) and on Windows I reckon that becomes `\r\n\r\n` so the PEG fails to match.

sogaiu:
Alternatively, may be it's just better to leave the `leap-year?` bits unexposed...sorry to be flipping back and forth :sweat_smile: 

Not really sure -- so may be that's another point in favor of not exposing for the moment?

CFiggers:
@bakpakin (No rush)

I don't have any other changes in the pipeline on this new module. If there's any updates you'd like to see please let me know.

Issue #249, opened 2025-10-03
Unexpected behavior of `(path/dirname ...)`


rwtolbert:
`dirname` leaves the trailing path separator on its result, which results in odd behavior if trying to use repeated calls to "climb" the directory tree.

```
repl:25:> (def p "/a/b/c/d")
"/a/b/c/d"
repl:26:> (path/dirname p)
"/a/b/c/"
repl:27:> (path/dirname (path/dirname p))
"/a/b/c/"
```

the trailing separator is not needed for subsequent calls to `path/join` 

```
repl:28:> (path/join (path/dirname p) "e")
"/a/b/c/e"
repl:29:> (path/join "/a/b/c" "e")
"/a/b/c/e"
```

For comparison, some Python output

```
>>> import os
>>> p = "/a/b/c/d"
>>> os.path.dirname(p)
'/a/b/c'
>>> os.path.dirname(os.path.dirname(p))
'/a/b'
```

If this existing code depends on this behavior of `dirname` perhaps a new method `(path/parent)` could be added that returned the parent directory without the trailing separator.

sogaiu:
> As for the linked examples, I don't see that behavior on the latest `spork/path`.

Nice -- looks like it may have been addressed in [this commit](https://github.com/janet-lang/spork/commit/c155231ff652089d5a76b655da9813eb10e37384).

rwtolbert:
Fixed in #256

PR #256, opened 2025-11-16
Fixes for Issue 249 - add `path/parent` as alternative to `path/dirname`

rwtolbert:
See #249 

This update adds a new `spork/path/parent` command that differs slightly from `dirname` in that it doesn't include the trailing slash and allows chained calls to climb up a directory tree. that is `(parent (parent "/foo/bar/baz"))` will return `/foo`.

While working on this on Windows, I added some updates to `path/abspath?` and `path/dirname` to include support for UNC paths like `\\wsl?\Ubuntu\home\foo`

so there are a number of new tests for all three methods, especially on Windows.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-latest, macos-latest, macos-13 ]
+        os: [ ubuntu-latest, macos-latest, macos-14 ]
         janet-version: [ heads/master, tags/v1.38.0, tags/v1.37.1, tags/v1.36.0, tags/v1.35.2 ]
     steps:
       - name: Checkout Janet


sogaiu:
Slowly going through here...I think digesting the Windows parts may take a while :sweat_smile: 

One thing about the `posix/parent` tests (haven't gotten much further yet):

```janet
(aeq (path/posix/parent "abc/def") "abc")
(aeq (path/posix/parent (path/posix/parent "/abc/def/xyz")) "/abc")
(aeq (path/posix/parent "abc/") "abc")
(aeq (path/posix/parent "abc") "")
(aeq (path/posix/parent "/") "/")
(aeq (path/posix/parent "./") ".")
```

May be it makes sense for some more obvious  case tos exist as tests too?  e.g.:

```janet
(aeq (path/posix/parent "/abc/def") "/abc")
(aeq (path/posix/parent "/abc/def/") "/abc/def")
```

I'm motivated by not just having edge cases but also to illustrate specific usages for folks new to things but also when folks who were once familiar return to things later.

Also, I think it makes it more explicit what type of thing is supported.

It's true that the second test (from what is already in the PR) seems to implicitly demonstrate at least one of the cases but may be it's not too expensive in this case to have a few more cases to make things more apparent?

WDYT?

rwtolbert:
@sogaiu Thanks for looking. I'm always happy to add more test cases.

sogaiu:
Thanks for using long-strings for the `win32/parent` tests, btw.  I find the tests much easier to read :+1:

Though perhaps it's worth changing these to match?

```janet
(aeq (path/win32/parent `c:\`) "c:\\")
(aeq (path/win32/parent `c:\foo`) "c:\\")
```

sogaiu:
I guess that in some of the proposed tests, the `basename` one here:

```janet
(aeq (path/win32/dirname "abc/") "abc/")
(aeq (path/win32/basename "abc") "abc")
(aeq (path/win32/dirname "abc") `.\`)
```

came about because of the already existing tests [here](https://github.com/janet-lang/spork/blob/e939a040f7183a518d9b396d017e1773e228a4c3/test/suite-path.janet#L47-L49C38):

```janet
(aeq (path/posix/dirname "abc/") "abc/")
(aeq (path/posix/basename "abc") "abc")
(aeq (path/posix/dirname "abc") "./")
```

May be it makes sense to change [the label of the section](https://github.com/janet-lang/spork/pull/256/files#diff-dcb3a418c1db239f17bb8db971f9ed2c3acaccd679a9755c7815991a9a750d09R59) to mention `basename` as well or separate out the `basename` tests?

I think the intent would be clearer which might help inform any future additions.

sogaiu:
I'm curious about the following because the expected values all end in backslashes but I had the impression that:

> leaves the trailing path separator on its result

was something we wanted to avoid, but I see in the tests:

```janet
(aeq (path/win32/parent `\\wsl$\Debian\foo`) `\\wsl$\Debian\`)
```

```janet
(aeq (path/win32/parent `\\wsl$\Debian\`) `\\wsl$\Debian\`)
```

```janet
(aeq (path/win32/parent `\\.\xyz\def`) `\\.\xyz\`)
```

These all look like [1] they end up with trailing separators...Are they supposed to?

---

[1] I verified that they do via running the code in the PR locally.

sogaiu:
If:

```janet
(aeq (path/win32/dirname `\\wsl$\Debian\foo`) `\\wsl$\Debian\`)
```

holds because `\\wsl$\Debian\` is a root directory [1] then I suppose:

```janet
(aeq (path/win32/dirname `\\wsl$\Debian\`) `\\wsl$\Debian\`)
```

is consistent with that...

But does the following [2] seem consistent with the first two?

```janet
(aeq (path/win32/dirname `\\wsl$\Debian`) `\\wsl$\Debian`)
```

Not sure how to think about these things :sweat_smile: 

Thanks for all the cases, btw :+1: 

---

[1] So in posix-land the analogy might be to `(dirname "/")` evaluating to `"/"`.

[2] Similar question for:

```janet
(aeq (path/win32/dirname `\\.\abc`) `\\.\abc`)
```

and:

```janet
(aeq (path/win32/dirname `c:`) `c:`)
```

sogaiu:
Thanks for the additional tests.

---

I've started to wonder about the `decl-win32-*` constructs.

Is it worth doing that over just directly defining:

* `win32/basename`
* `win32/dirname`
* `win32/parent`
* `win32/parts`

?

May be @bakpakin has an opinion about this.

sogaiu:
> and is the conclusion, we just need to remove the old macos-13 tests?

Not sure.

[Some related material](https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/):

> The macOS 13 hosted runner image is closing down, following our [N-1 OS support policy](https://github.com/actions/runner-images?tab=readme-ov-file#software-and-image-support). This process will begin October 1, 2025, and the image will be fully retired on December 4, 2025. We recommend updating workflows to use macos-14 or macos-15.

May be it's just [this line](https://github.com/janet-lang/spork/blob/e939a040f7183a518d9b396d017e1773e228a4c3/.github/workflows/test.yml#L15) that is relevant?

So possibly replacing `macos-13` with something else...

sogaiu:
> Thanks for pointing that out. I'm still learning how to work on Actions.

I'm no expert :)

> Looks like the best thing is to change that to "macOS-14" to support the oldest GitHub will allow.

That's how it seems to me too.

sogaiu:
May be it's better to make a separate PR for the CI change?

PR #259, opened 2025-11-26
CI: Add 1.39.1 and remove 1.35.2 testing

sogaiu:
This PR is an attempt to update the versions of janet that are tested against in CI.

The intent is to add testing for 1.39.1 and stop testing 1.35.2.


pepe (approved):
LGTM

PR #260, opened 2025-12-01
getline: Add support for ctrl-u

tw4452852:
Don't know the original reason of not having this shortcut, but it's nice to have, isn't it?

Change-Id: I1c35f486ca5e3e1fb02a607572583a896a6a6964

### DIFF
--- a/spork/getline.janet
+++ b/spork/getline.janet
@@ -350,6 +350,8 @@
             (set hindex (history-move hindex 1))
             17 # ctrl-q
             (do (set more-input false) (set ret-value :cancel) (clear-lines))
+            21 # ctrl-u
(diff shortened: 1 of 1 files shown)


tw4452852:
> I guess this is useful if one does not want to eliminate the whole line, but just from where the cursor is back to the start.

Yes, this is what I found when I implemented a shell with `getline`.

> On a separate note, I noticed that ctrl-a and ctrl-u position my cursor one position to the right of the beginning of the line. Does that happen for other folks?

I test it locally, it works as expected:

Before:
screenshot-20251201@181818

After pressing `<ctrl+u>`:
screenshot-20251201@181921

tw4452852:
Ping

PR #262, opened 2025-12-02
getline: bugfix for <ctrl-a>

tw4452852:
According to https://unix.stackexchange.com/a/559331, moving zero column is implemented inconsistently between different terminal emulators.

In order to fix this inconsistency, we avoid using this escape sequence entirely.

Change-Id: I9f6358680a6e054c2846b823af8e265c6a6a6964


sogaiu:
I gave this a try and I now observe the desired behavior :+1:

PR #264, opened 2025-12-11
pm: satisfy flycheck

tw4452852:
Original issue:

```
/home/tw/code/spork > janet -k spork/pm.janet
error: spork/pm.janet:419:1: compile error: (macro) grammar error in nil, unexpected peg source
  in peg/match [src/core/peg.c] on line 1924
  in make-template [spork/pm.janet] on line 384, column 14
  in deftemplate [spork/pm.janet] on line 406, column 45
  in dofile [boot.janet] (tail call) on line 3132, column 7

```

Change-Id: I54f8c197323aa10ea6111f6eb83b86e16a6a6964

### DIFF
--- a/spork/pm.janet
+++ b/spork/pm.janet
@@ -370,7 +370,7 @@
 ### Generate new projects quickly, ported from jpm
 ###
 
-(def- template-peg
+(def- template-peg :flycheck
   "Extract string pieces to generate a templating function"
   (peg/compile
     ~{:sub (group


sogaiu:
Possibly of interest: https://github.com/janet-lang/jpm/issues/105

---

One issue of using `*flychecking*` is that IIUC [it was only added in 1.40.0](https://github.com/janet-lang/janet/blob/efe66f207f94e194c25c76a312e16dcdf160c50b/CHANGELOG.md#1400---2025-11-15) -- so I think the proposed change won't have the desired effect for versions of `janet` early than 1.40.0...but may be that doesn't matter.

FWIW, [here](https://github.com/janet-lang/spork/blob/6e14b4d907dedd51d8448c67da33572c5a0f797f/.github/workflows/test.yml#L16) is an indication of versions of `janet` that spork is tested against (though I don't know that this is a "policy").

tw4452852:
> One issue of using *flychecking* is that IIUC [it was only added in 1.40.0](https://github.com/janet-lang/janet/blob/efe66f207f94e194c25c76a312e16dcdf160c50b/CHANGELOG.md#1400---2025-11-15)

Yep, I'm thinking about if there's a compatible way to handle this.

tw4452852:
@sogaiu 
Ready for review.

tw4452852:
> For 1.39.1 and below I get something like:
> 
> error: /home/user/src/spork/spork/pm.janet:420:1: compile error: (macro) grammar error in nil, unexpected peg source
>   in peg/match [src/core/peg.c] on line 1904
>   in make-template [/home/user/src/spork/spork/pm.janet] on line 385, column 46
>   in deftemplate [/home/user/src/spork/spork/pm.janet] on line 407, column 45
>   in dofile [boot.janet] (tail call) on line 3060, column 7

Yeah, this is expected.

> One minor thing -- may be it makes sense to use @[] instead of [] here:

Good catch, I've updated the change.

tw4452852:
Ah, definitely yes, `*flychecking*` is just an alias to `:flychecking`.

After a second thought, if https://github.com/janet-lang/janet/pull/1681 could be accepted, adding `:flycheck` on `template-peg` would be a better option.

PR #285, opened 2026-04-05
Address #284 - handle path/normalize edge case

sogaiu:
This PR is an attempt to address #284.

`path/normalize` builds up `accum`, a stack of path segments, which is ultimately used as part of constructing a final path.

The conditions under which `accum` is pushed to and popped from differ slightly depending on whether the original path is absolute or not.

The proposed changes attempt to distinguish the cases and act accordingly.

Also included are some regression tests for a couple of cases, expressed for the posix and win32 cases separately.

### DIFF
--- a/spork/path.janet
+++ b/spork/path.janet
@@ -116,9 +116,11 @@
        (match x
          [:lead what] (set lead what)
          "." nil
-         ".." (if (= 0 seen)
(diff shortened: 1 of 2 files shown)


bakpakin:
Thanks! LGTM

PR #290, opened 2026-04-20
getline: respect NO_COLOR when showing doc

tw4452852:


sogaiu:
I think there is also [a dynamic variable `*doc-color*` which `doc-format` respects](https://github.com/janet-lang/janet/blob/c6b802b082aa21b6e65e8c2729c85237b0cc8a24/src/boot/boot.janet#L3318-L3320).

May be particular use cases can set `*doc-color*` via the `NO_COLOR` environment variable and the last parameter to the call here to `doc-format` can be removed?

tw4452852:
Ah, haven't noticed this exists.

However, this reminds me what if we handle `NO_COLOR` inside `doc-format` itself. Then all the callsites have the benefit automatically unless they force coloring with the parameter . What do you think?

sogaiu:
I don't have a clear idea of what would be good.  Some thoughts follow:

* In the context of `boot.janet`, things might be redundant because `NO_COLOR` handling happens already elsewhere via [a call to `apply-color`](https://github.com/janet-lang/janet/blob/c6b802b082aa21b6e65e8c2729c85237b0cc8a24/src/boot/boot.janet#L4758-L4760).  I guess the dynamic variable's value should take precedence?
* The default behavior of `default-doc-fetch` is being changed.  I don't know if anyone already relies on the previous behavior.  I have had things break because of changes in color behavior elsewhere though and it's not something I wish on others.
* `doc-format`'s behavior becoming dependent directly on something external seems like a recipe for increasing unexpected behavior.

Hmmm...may be these are either non-issues or not really important?  I guess it seems to me that it's not a clear win to change things, but the above are just initial thoughts.

It's not my call though in any case :sweat_smile:

PR #293, opened 2026-06-12
argparse: handle `--param=5` just like `--param 5`

### LINKED ISSUE #292: argparse suggestion: handle `--thing=5` as well as `--thing 5`
Traditionally, UNIX tools that take option parameters with values have allowed passing the value either as a second parameter (like `--input-file somefile.txt`) or with an equal sign separating the option name from the value (like `--input-file=somefile.txt`). While there's no formal standard, the [`getopt_long` function](https://man7.org/linux/man-pages/man3/getopt_long.3.html) is the de-facto standard, and it allows long options to take parameters in either form, both space-separated or equals-separated.

Currently `argparse` can only handle the space-separated form. Trying `./example.janet -k foo --thing=5` produces the error "usage error: unknown option thing=5". It would be nice if argparse could handle the `--thing=5` syntax as well as the `--thing 5` syntax.

The implementation could work like this: if argparse looks up a long option name and it's not found, instead of immediately calling `(usage "unknown option " name)`, it could instead have one more check. If the name contains at least one equals sign and the part before the **first** equals sign (in a minute you'll see why it's important that it be the first one) is a valid option name, then the name is split on the first equals sign into a name and a value.

The reason why it's important for it to be the first equals sign is in case you have a parameter like `--env` that expects to take parameters that have the shape `KEY=VALUE` themselves. Depending on the shell that's processing command lines before Janet sees them, it's sometimes tricky to write `--env KEY=VALUE` and have the string `KEY=VALUE` get passed in to the program; the shell might interpret that as setting an environment variable for the shell, rather than as the intended text of a parameter. And so some people might be forced to try `--env=KEY=VALUE`; it would be nice if `argparse` handled that correctly, rather than erroring out saying "Too many = characters" or "unknown option env=KEY".

This is only a "would be nice to have" feature, not a "can't live without it" feature. But it would be fairly straightforward to implement. (The syntax for short parameters with values that `getopt` allows, where `-k foo` can also be written as `-kfoo`, would be a lot trickier to handle, and I'm not asking for that one).

rmunn:
This copies `getopt_long` behavior, where either space or equals are valid separators between options and values.

This PR does not allow `-p=5` since `getopt` doesn't allow equals signs in short options (at least, that's not documented). Passing values to short options still requires spaces. If equals signs between option and value are desirable in short options, it would be quite easy to add that later.

Fixes #292.

### DIFF
--- a/spork/argparse.janet
+++ b/spork/argparse.janet
@@ -130,7 +130,7 @@
 
   # Handle an option
   (defn handle-option
-    [name handler]
+    [name handler &opt value]
     (array/push (res :order) name)
     (case (handler :kind)
       :flag (put res name true)
@@ -138,16 +138,16 @@
                (var counter (or (get res name) 0))
                (++ counter)
                (put res name counter))
-      :option (if-let [arg (get args i)]
+      :option (if-let [arg (or value (get args i))]
                 (do
                   (put res name arg)
-                  (++ i))
+                  (unless value (++ i)))
                 (usage "missing argument for " name))
-      :accumulate (if-let [arg (get args i)]
+      :accumulate (if-let [arg (or value (get args i))]
                     (do
                       (def arr (or (get res name) @[]))
                       (array/push arr arg)
-                      (++ i)
+                      (unless value (++ i))
                       (put res name arr))
                     (usage "missing argument for " name))
       # default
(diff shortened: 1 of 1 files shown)


bakpakin:
This stagnated a while since the windows CI was failing, but that has been fixed on main and so can be ignored. LGTM, merging!

Issue #294, opened 2026-06-15
http does not accept responses without a Reason Phrase


birchb1024:
Some HTTP servers (like mine) do not bother to respond with `200 OK`, just `200 ` (trailing space)

When doing a POST with spork/http like this

```janet
(defn table-to-form-encoded [data]
  (string/join (map
                 (fn [p]
                   (let 
                    [k (p 0)
                     v (p 1)
                     fmt (cond  (integer? v) "d" (number? v)  "g" (string? v)  "s" true "m")]
                     (string/format (string "%s=%" fmt) k v))) 
                 (pairs data)) "&"))

(defn make-post-request [url data]
  (let [body (table-to-form-encoded data) headers {"Content-Type" "application/x-www-form-urlencoded"}]
    (printf "%m" ['make-post-request headers body])
    (let [response (http/request "POST" url :headers headers :body body)]
      (printf "25 %m" response)
      (print "Status Code: " (response :status))
      (print "Response Body: " (response :body)))))
```
The POST is successful in that it's accepted by the server, but `spork/request` just throws an `error` . 

However the RFC says that the Reason Phrase is optional. Refer [StackOverflow](https://stackoverflow.com/questions/17517086/can-an-http-response-omit-the-reason-phrase)

Please consider changing `(some :printable)` to `(any :printable) in `http.janet` :

```janet
(def- http-grammar                                                                        
  ~{:request-status (* :method :ws :path :ws "HTTP/1." :d :any-ws :rn)
    :response-status (* "HTTP/1." :d :ws (/ ':d+ ,scan-number)
                        :ws '(some :printable) :rn)
```

sogaiu:
Optionality seems consistent with what can be seen [here](https://www.rfc-editor.org/info/rfc9112/#name-status-line):

> The first line of a response message is the status-line, consisting of the protocol version, a space (SP), the status code, and another space and ending with an OPTIONAL textual phrase describing the status code.

The changes seems like a reasonable one to me (ensuring that there is whitespace after the response code as mentioned).

May be a PR would be welcome.

birchb1024:
Made the code and tst changes, 
what do I need to do to test this and make a PR? 

The instructions say "Spork must be installed to test properly.", but it fails. 

```
$ janet --install .                                                                                                   
error: bundle "spork" is already installed
  in errorf [boot.janet] on line 207, column 3
  in bundle/install [boot.janet] on line 4445, column 5
[...]

PR #295, opened 2026-06-16
Allow empty reason phrase in http grammar, The RFC says that the Reas…

birchb1024:
Allow empty reason phrase in http grammar, The RFC says that the Reason Phrase is optional.

Refer StackOverflow https://stackoverflow.com/questions/17517086/can-an-http-response-omit-the-reason-phrase

See isse #294


sogaiu:
I tried the proposed changes and they worked here.

pepe (approved):
LGTM

PR #297, opened 2026-07-22
Print test output to stderr

pepe:
## Summary

Write `spork/test` suite summaries to stderr instead of stdout and update the stream-sensitive test accordingly.

## Motivation

Assertion diagnostics already use stderr, but `end-suite` still wrote its human-facing summary to stdout. This prevented callers from reserving stdout for machine-readable subprocess protocols. In my methodology, I use stdout as the interprocess communication channel.

This change makes ordinary test reporting consistently use stderr while leaving the documented stdout behavior of `timeit` and `timeit-loop` unchanged.

## Impact

Visible test output remains the same in a terminal. Programs capturing stdout will no longer receive the final suite summary.

## Validation

- Verified `capture-stderr` receives the complete suite summary.
- Verified a representative local test process exits successfully with:
  - stdout: 0 bytes
  - stderr: the suite summary
- Rebased cleanly onto current `origin/master`.

The summary was written and revised with the help of an LLM.

### DIFF
--- a/spork/test.janet
+++ b/spork/test.janet
@@ -75,8 +75,8 @@
   "Ends test suite, prints summary and exits if any tests have failed."
   []
   (def delta (- (os/clock) start-time))
-  (prinf "test suite %V finished in %.3f seconds - " suite-num delta)
-  (print num-tests-passed " of " num-tests-run " tests passed.")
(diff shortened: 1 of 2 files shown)


bakpakin:
Not sure why windows CI is failing, merging anyway

PR #300, opened 2026-08-10
Non-ASCII characters in assert text

dmitrys99:
## What's included

I made test library, but instead, say, `"¡Hola!"`, I see `"\xC2\xA1Hola!"` for `(assert c "¡Hola!")`.

Root cause: for string `(describe ...)` call outputs hex-encoded byte sequence, not the string itself.

## Fix
Use `(string? ...)` predicate and return string for description.

### DIFF
--- a/spork/test.janet
+++ b/spork/test.janet
@@ -23,7 +23,10 @@
   (def line-info (string/format "%s:%d"
                                 (frame :source) (frame :source-line)))
   (if x
-    (when (is-verbose) (eprintf "\e[32m✔\e[0m %s: %s: %v" line-info (describe e) x))
+    (when (is-verbose) (eprintf "\e[32m✔\e[0m %s: %s: %v"
+                                line-info
+                                (if (string? e) e (describe e))
+                                x))
     (do
       (eprintf "\e[31m✘\e[0m %s: %s: %v" line-info (describe e) x) (eflush)))
   x)
--- a/spork/test.janet
+++ b/spork/test.janet
@@ -18,16 +18,16 @@
     (break x))
   (default e "assert error")
   (when x (++ num-tests-passed))
-  (def stack (debug/stack (fiber/current)))
-  (def frame (last stack))
-  (def line-info (string/format "%s:%d"
-                                (frame :source) (frame :source-line)))
-  (def desc (if (string? e) e (describe e)))
-
+  (defn line-data []
+    (def stack (debug/stack (fiber/current)))
+    (def frame (last stack))
+    (string/format "%s:%d: %s: %v"
+                   (frame :source) (frame :source-line)
+                   (if (string? e) e (describe e)) x))
   (if x
-    (when (is-verbose) (eprintf "\e[32m✔\e[0m %s: %s: %v" line-info desc x))
+    (when (is-verbose) (eprintf "\e[32m✔\e[0m %s" (line-data)))
     (do
-      (eprintf "\e[31m✘\e[0m %s: %s: %v" line-info desc x) (eflush)))
+      (eprintf "\e[31m✘\e[0m %s" (line-data)) (eflush)))
   x)
 
 (defn skip-asserts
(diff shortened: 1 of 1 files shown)


sogaiu:
It seems that `describe` started to be used in [this commit](https://github.com/janet-lang/spork/commit/fd761c3732f9969b639836064d3f6ccf368a3230).

@bakpakin Do you remember if there were specific reasons for that?

May be its output can be more informative?  It looks like `describe`'s current docstring is:

```text
Returns a string that is a human-readable description of x. For 
recursive data structures, the string returned contains a pointer 
value from which the identity of x can be determined.
```

sogaiu:
In any case, if the change makes sense to apply, may be [two lines down](https://github.com/janet-lang/spork/blob/11b335f7b5159fcdf2f595d9636b1eeeb87eb867/spork/test.janet#L28) ought to be modified as well?

```janet
(eprintf "\e[31m✘\e[0m %s: %s: %v" line-info (describe e) x) (eflush)))
```

kamisori:
> I made test library, but instead, say, `"¡Hola!"`, I see `"\xC2\xA1Hola!"` for `(assert c 
"¡Hola!")`.

what terminal are you using? 

for foo.janet :=

```
(pp (describe "¡Hola!"))
(pp (string "¡Hola!"))
(pp "¡Hola!")
(print "¡Hola!")
(assert nil "¡Hola!")
```

i see the expected output of

```
$ janet foo.janet 
"\"\\xC2\\xA1Hola!\""
"\xC2\xA1Hola!"
"\xC2\xA1Hola!"
¡Hola!
error: ¡Hola!
  in thunk [foo.janet] (tail call) on line 5, column 1

$ uname -a
Linux asdf 6.8.0-137-generic #137-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul 17 20:28:23 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux
```

sogaiu:
Possibly obvious to all, but just to be clear, [`assert` in spork's `test` has been overridden](https://github.com/janet-lang/spork/blob/11b335f7b5159fcdf2f595d9636b1eeeb87eb867/spork/test.janet#L37-L45) and is not the same as the built-in `assert`.

dmitrys99:
> (assert nil "¡Hola!")

You're using core `(assert)`, the problem is with `(spork/test/assert)`.

> In any case, if the change makes sense to apply, may be [two lines down](https://github.com/janet-lang/spork/blob/11b335f7b5159fcdf2f595d9636b1eeeb87eb867/spork/test.janet#L28) ought to be modified as well?

Done

sogaiu:
~May be one of the branches is missing an `(eflush)`?~

~The [commit where `assert-no-tail` was added](https://github.com/janet-lang/spork/commit/fd761c3732f9969b639836064d3f6ccf368a3230) only had `(eflush)` in one place, but perhaps a second one was intended (or perhaps there should be none in either branch).~

_Update_: NVM, the first branch likely doesn't need one (because execution isn't about to halt) and can have better performance from buffering.  We want the call in the second branch to ensure we get the output (aborting due to error might result in information loss).  Or something like that :)

sogaiu:
~I will make a separate issue about `eflush`, but if anyone has something to say about it, I'd be happy to hear :)~

sogaiu:
At present the PR uses `string?` to pass `e` through without it being processed by `describe`.

I think this means buffers, keywords, and symbols will end up being passed through `describe` and be harder to perceive (presumably in a similar manner to what motivated this PR).  Perhaps it's unlikely that `e` will have any of these types though...may be it's just better to keep things simple and wait to see if any issues arise.

bakpakin:
> @bakpakin Do you remember if there were specific reasons for that?

I believe I was getting non-ascii output from a test suite that was confusing to read. I'm assuming that has since been resolved and is certainly confusing.

PR #302, opened 2026-08-24
Implement `utf8/is-valid?`

gahr:
Hi, I'm new to Janet and liking it very much so far!

I need to check if a string contains valid UTF-8.
I looked around but couldn't find anything. So here's my attempt.

Nothing too clever here, I'm just exposing the bulk of `decode-rune` as a helper function, and looping over the buffer.

### DIFF
--- a/src/utf8.c
+++ b/src/utf8.c
@@ -26,18 +26,7 @@
  * See https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt for a list
  * of test cases. */
 
-JANET_FN(cfun_utf8_decode_rune,
-        "(utf8/decode-rune buf &opt start)",
-        "Read a UTF-8 encoded Unicode codepoint from the buffer which starts at the given index. Returns a tuple [value width], where width = number of bytes consumed. If at the end of buffer or the buffer contains malformed UTF-8, returns [nil 0].") {
-    janet_arity(argc, 1, 2);
-    JanetByteView buf = janet_getbytes(argv, 0);
-    int32_t start;
-    if (argc > 1) {
-        start = janet_getinteger(argv, 1);
-    } else {
-        start = 0;
-    }
-
+static JanetTuple decode_rune(JanetByteView buf, int32_t start) {
     Janet res[2] = { janet_wrap_nil(), janet_wrap_integer(0) };
     if (start >= buf.len) {
         goto exit;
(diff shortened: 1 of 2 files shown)


sogaiu:
Hello :)

I gave it a look and a try.

Seems ok.

A few comments:

1. How about some tests for things that should pass as well?  For example:

    ```janet
    (assert (true? (utf8/is-valid? "hello")) "utf8: is-valid? on ascii")
    (assert (true? (utf8/is-valid? "안녕")) "utf8: is-valid? on valid non-ascii")
    ```

2. A part of the commit message "body" seems a bit out of place -- it seems like it's the content of the first comment in this PR.  How about something close to the last sentence?

    > Exposes the bulk of decode-rune as a helper function, and looping over the buffer.

gahr:
>     2. A part of the commit message "body" seems a bit out of place -- it seems like it's the content of the first comment in this PR.  How about something close to the last sentence?
>        > Exposes the bulk of decode-rune as a helper function, and looping over the buffer.

ehm, yeah, the PR description comes from the commit message. I'm fine with removing that part.

gahr:
I segregated the `utf8/is-valid?` tests into their own section, added tests for valid cases, and reworked the commit message.

sogaiu:
Looks pretty good.

BTW, if you use `janet-format` [1] on `test/suite-utf8.janet`:

```
janet-format test/suite-uft8.janet
```

it may tidy things up a bit.

---

[1] May be you know, but FWIW it's a script that's part of a spork install.

gahr:
> BTW, if you use `janet-format` [1] on `test/suite-utf8.janet`:

Thanks, I wasn't aware of the tool (still very in my early days in Janet). Applied.

sogaiu:
LGTM

bakpakin:
```c
   while (i < buf.len) {
        JanetTuple res = decode_rune(buf, i);
        if (janet_checktype(res[0], JANET_NIL))
            return janet_wrap_boolean(0);
        i += janet_unwrap_integer(res[1]);
    }
```


Why are we allocating a tuple for every character? This is incredibly inefficient. We should certainly have a routine that does not do this allocation if we are going to provide a C function to make this check faster.

bakpakin:
Did not mean to close this, but we really shouldn't be allocating any memory for this routine.

gahr:
> Did not mean to close this, but we really shouldn't be allocating any memory for this routine.

Definitely! I have amended the code.

bakpakin:
Looks much better

PR #304, opened 2026-08-26
Make sure to produce tuple literals

### LINKED ISSUE #303: janet-pm fails to build a project depending on jsec
I'm running janet 1.41.2 on FreeBSD 15.0-RELEASE, with spork installed from the master branch (as of https://github.com/janet-lang/spork/commit/3918802d6b79848a3dba113b1fe2ee1a8f7b667b) with:

```sh
$ janet -l ./bundle -e '(build)'
$ doas janet --install .
```

Given this `project.janet`

```janet
(declare-project
  :name "foo"
  :description ```foo ```
  :author ```gahr ```
  :dependencies @["spork" "https://github.com/llmII/jsec" ]
  :version "0.0.0")

(declare-executable
  :name "foo"
  :entry "foo/init.janet")
```
and this `foo/init.janet`

```janet
(import jsec/tls :as tls)
```

building the project (after successfully installing the dependencies) fails:

```
$ /usr/local/lib/janet/bin/janet-pm build
generating executable c source _build/develop/foo.c from foo/init.janet...
found native /usr/local/lib/janet/jsec/dtls-stream.so (/usr/local/lib/janet/jsec/dtls-stream.so)...
found native /usr/local/lib/janet/jsec/tls-stream.so (/usr/local/lib/janet/jsec/tls-stream.so)...
found native /usr/local/lib/janet/jsec/utils.so (/usr/local/lib/janet/jsec/utils.so)...
rule _build/develop/foo failed: error: expected integer key for string in range [0, 5), got "-lcrypto"
  in thunk [eval] (tail call) on line 4, column 11
  in eval-string [boot.janet] on line 2832, column 36
  in target [/usr/local/lib/janet/spork/declare-cc.janet] on line 800, column 21
  in try [/usr/local/lib/janet/spork/build-rules.janet] on line 130, column 26
error: failed to build
  in run-rules [/usr/local/lib/janet/spork/build-rules.janet] (tail call) on line 162, column 5
  in build [/usr/local/lib/janet/spork/declare-cc.janet] (tail call) on line 320, column 5
  in run-main [boot.janet] on line 4684, column 16
  in cli-main [boot.janet] on line 4904, column 17
```

gahr:
Fixes #303

I'm not sure about `static-libs` and `smart-libs`.

### DIFF
--- a/spork/declare-cc.janet
+++ b/spork/declare-cc.janet
@@ -472,6 +472,8 @@
   (def rules (get-rules))
 
   # Defaults
+  (default lflags @[])
+  (default ldflags @[])
   (default libs (thaw ldflags))
   (default embedded @[])
   (default nostatic false)
@@ -585,8 +587,8 @@
               (string name asuffix)
               {:static-entry ename
                :cpp has-cpp
-               :ldflags (if libs ~[,;libs] '[])
-               :lflags (if lflags ~[,;lflags] '[])
+               :ldflags ~[,;libs]
+               :lflags ~[,;lflags]
                :static-libs static-libs
                :smart-libs smart-libs
                :use-rdynamic use-rdynamic
--- a/spork/declare-cc.janet
+++ b/spork/declare-cc.janet
@@ -587,7 +587,7 @@
                :cpp has-cpp
                :ldflags (if libs ~[,;libs])
                :lflags (if lflags ~[,;lflags])
-               :static-libs static-libs
+               :static-libs (if static-libs ~[,;static-libs])
                :smart-libs smart-libs
                :use-rdynamic use-rdynamic
                :use-rpath use-rpath})))
--- a/spork/declare-cc.janet
+++ b/spork/declare-cc.janet
@@ -472,8 +472,6 @@
   (def rules (get-rules))
 
   # Defaults
-  (default lflags @[])
-  (default ldflags @[])
   (default libs (thaw ldflags))
   (default embedded @[])
   (default nostatic false)
@@ -587,8 +585,8 @@
               (string name asuffix)
               {:static-entry ename
                :cpp has-cpp
-               :ldflags ~[,;libs]
-               :lflags ~[,;lflags]
+               :ldflags (if libs ~[,;libs])
+               :lflags (if lflags ~[,;lflags])
                :static-libs static-libs
                :smart-libs smart-libs
                :use-rdynamic use-rdynamic
--- a/spork/declare-cc.janet
+++ b/spork/declare-cc.janet
@@ -585,8 +585,8 @@
               (string name asuffix)
               {:static-entry ename
                :cpp has-cpp
-               :ldflags libs
-               :lflags lflags
+               :ldflags (if libs ~[,;libs] '[])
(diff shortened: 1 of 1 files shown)


sogaiu:
> I'm not sure about static-libs and smart-libs.

I looked into these and found:

* [`*static-libs*`](https://github.com/janet-lang/spork/blob/3918802d6b79848a3dba113b1fe2ee1a8f7b667b/spork/cc.janet#L43)
* [`*smart-libs*`](https://github.com/janet-lang/spork/blob/3918802d6b79848a3dba113b1fe2ee1a8f7b667b/spork/cc.janet#L52-L54)

---

Since the bit for `*static-libs*` mentions:

> List of static libraries to use when compiling

this one seems like it could use a similar treatment as for `lflags` and `ldflags`.

---

For `*smart-libs*` it looks like:

> Try to resolve circular or out-of-order dependencies between libraries by using --start-group and --end-group.
> Some linkers support this by default, but not at all. Defaults to true on linux and macos.

So I don't think anything needs to be done for it.

May be the name `*use-smart-libs*` would have been less confusing but perhaps that ship has sailed :)

sogaiu:
Regarding:

> Deal with nil as soon as possible

I'm not sure how the corresponding values for `:ldflags` and `:lflags` get used once the associated `.meta.janet` file is processed.

However, given that in [`jpm`'s version](https://github.com/janet-lang/jpm/blob/04603ffe61f948a28307980ebf7d6f09ba4798ec/jpm/declare.janet#L122-L123) there is:

```janet
                          :ldflags ~',(opts :ldflags)
                          :lflags ~',(opts :lflags)})))
```

...may be a `nil` value for `ldflags` / `lflags` is ok [1] or even expected.

Not sure though :sweat_smile: 

---

[1] Since in Janet, a `nil` as a value in a table or struct will cause the corresponding "entry" to be removed.

gahr:
Thanks @sogaiu , sure, it makes sense to keep nils, so they don't end up in the final table. I have amended the code.

If you prefer, I can keep the code more similar to what jpm does, `:lflags ~',lflags` but that'll produce `:lflags (quote ("-lssl" "-lcrypto"))`, which is probably not as nice.

sogaiu:
I agree it's nicer the way you've done it.  I think it should effectively lead to similar behavior as the `jpm` version so I think it's fine.  Hopefully that's correct, if not, may be another reviewer will spot it :sweat_smile: 

---

I think investigations in [this earlier comment](https://github.com/janet-lang/spork/pull/304#issuecomment-5434108594) suggest that something similar should be done for `*static-libs*` (though not for `*smart-libs*`).  Sorry if that was unclear.

---

Hmm, strange...when I look at [this](https://github.com/janet-lang/spork/pull/304/changes) (the "Files changed" tab near the top of the page), I see `src/utf8.c` and `test/suite-utf8.janet` listed too...is that buggy behavior on GH's part?

gahr:
hah sorry I merged my other feature branch cause I needed both features and I pushed it. Fixed.

sogaiu:
Thanks for the update and change, LGTM.

Issue #291, opened 2026-05-01
Handling of keep-alive connections in `spork/http`


yohannd1:
Hello! This is more of a question than anything at the moment (should this be here? Not sure if it would fit in the Discussions tab over at the janet repo).

I'm porting a script from Python to Janet which involves making a HTTP server. In this script, one of the features is a special `/api/stream` route(?) which replies the client with a continuous event stream. The logic is, superficially:
```python
self.send_response(200)
self.send_header("Content-Type", "text/event-stream")
self.send_header("Cache-Control", "no-cache")
self.send_header("Connection", "keep-alive")
self.end_headers()

wait_for_event:
    self.send_utf8("message_text\n\n")
    self.flush()
```

I'd like to know if there currently is any way of doing that with `spork/http`, as a struct is returned on request results. I have noticed that the `:body` field of said struct can be an iterable, but I'm struggling to figure out how to adapt a function to an iterable.

bakpakin:
You should be able to use an iterable like so:

```janet
(def chunks
  (coro
    (for i 0 10
      (ev/sleep 0.1) # do some work, take from a channel with ev/take, or wait for an event
      (yield (string/format "chunk %d" i)))))
```

And pass `chunks` to :body where that is expected. Does that help?

yohannd1:
Oh, had forgotten about `yield`! Thanks. Ended up using this with a channel, indeed.
```janet
(defn reply-evstream []
  (def headers
    {"Content-Type" "text/event-stream"
     "Cache-Control" "no-cache"
     "Connection" "keep-alive"})

  (def body-fiber
    (coro
      (while (def x (ev/take *reload-queue*))
        (yield "data: reload\n\n"))))

  {:status 200 :headers headers :body body-fiber})
```

yohannd1:
(Not sure if it's the best idea to reopen this, but it's very related)
Hello again! I've been using this and it works fine, except for the fact that closed connections are erroring out with `Broken pipe` errors - that's to be expected, but I want to catch it to at least try and silence it (so it doesn't pollute the CLI output). Is there any way to do that? I couldn't think of one since the fiber yields values to be sent.

PR #306, opened 2026-09-13
Allow error handling in http/send-response

yohannd1:
An attempt to tackle #291. Consists of adding an optional callback function to the response dictionary `http/send-response`, `:error`, which if specified handles errors on sending the response.


bakpakin:
Any reason wrapping `http/send-response` or `http/request` in an error handler doesn't work here?

bakpakin:
For some more context, the sigpipe errors will only happen when running HTTP over a non-socket stream, like a pipe. The reason for this is that `(:write stream buf)` will use `send` internally for sockets with MSG_NOSIGNAL, but `write` for all other file descriptors. In an HTTP server, ideally we would handle EPIPE by simply closing the connection and not even bother to log an error (or perhaps you might want to log something?).

yohannd1:
I had put the wrapping in `write-body` because for some reason I thought multiple places would make use of it, but looking back now, only really `http/send-response` does. I don't think `http/request` needs it because, at least as far as I know, it's not called in a server handler (or the sort) (do correct me if I'm wrong!).

As for handling EPIPE, I'm not sure myself - I was thinking of a more general solution in case some other sort of error appears, though I myself can't think of many cases like that (maybe an error while going through the iterable, in case it's a coroutine?). Plus I do wonder if some applications would benefit from logging "connection lost" errors (mine doesn't).

yohannd1:
Pushed a fix for Windows (hopefully should work, I can't test locally). I'm not fully sure if this is the best way to detect whether the system is POSIX-ish, though.

PR #309, opened 2026-09-19
feat(http): streaming client with open-stream, :blocks, and RFC 9110 …

llmII:
Introduce pure Janet streaming HTTP client capabilities to spork/http:
- Add http/open-stream returning a stream-like object supporting:
  - :read (byte count, :line, :all)
  - :blocks (fiber yielding buffer slices)
  - :lines (fiber yielding clean delimited lines)
  - :close
- Add http/download to stream responses directly into files, buffers, or custom sinks without buffering entire responses
- Add http/resolve-url for RFC 9110 relative URL resolution
- Add http/make-response-stream for constructing stream-like response objects
- Update http/request with :stream option and :sink streaming support
- Update http/read-body with optional sink streaming support
- Use compiled PEGs for RFC 9112 chunk header, chunk terminator, and line parsing
- Implement transparent HTTP chunked transfer-encoding decoding
- Implement automatic RFC 9110 redirect following (301, 302, 303, 307, 308)
- Add comprehensive test coverage in test/suite-http.janet


bakpakin:
Looks like a lot of functionality! I do wonder about the performance delta though, I see a lot of use of possibly slower functionality, such as the default `request` now making a lot more allocations and such even when the new behavior is not used. I suppose pure Janet http was never winning on any benchmarks but curious on the perf impact.

